PR #23, opened 2019-04-01
pkg/providers: Adding HashiCorp Vault provider

anubhavmishra:
This PR adds the initial version of the HashiCorp Vault provider for Secret Store CSI driver.


pkg/providers/vault/examples/nginx-pod-keyvault.yaml
```diff
@@ -0,0 +1,16 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: nginx-keyvault
```
ritazh: Should we rename these to `-vault` instead of `-keyvault` to distinguish this example from the Azure Key Vault one.

pkg/providers/vault/examples/nginx-pod-keyvault.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  containers:
+  - image: nginx
+    name: nginx-keyvault
```
ritazh: nit: rename

pkg/providers/vault/examples/nginx-pod-keyvault.yaml
```diff
@@ -0,0 +1,16 @@
...
+    volumeMounts:
+    - name: keyvault01
```
ritazh: nit: rename

pkg/providers/vault/examples/pv-keyvault-csi.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: pv-keyvault
+spec:
+  capacity:
+    storage: 1Gi
+  accessModes:
+    - ReadOnlyMany
+  persistentVolumeReclaimPolicy: Retain
+  csi:
+    driver: keyvault.csi.k8s.com
```
ritazh:
```suggestion
    driver: secrets-store.csi.k8s.com
```

https://github.com/deislabs/secrets-store-csi-driver/blob/master/deploy/example/pv-secrets-store-csi.yaml#L12

pkg/providers/vault/examples/pvc-keyvault-csi-static.yaml
```diff
@@ -0,0 +1,12 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: pvc-keyvault
```
ritazh: nit: rename

pkg/providers/vault/examples/pvc-keyvault-csi-static.yaml
```diff
@@ -0,0 +1,12 @@
...
+spec:
+  accessModes:
+    - ReadOnlyMany
+  resources:
+    requests:
+      storage: 1Gi
+  volumeName: pv-keyvault
```
ritazh: nit: rename

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+)
+
+const (
+	// VaultObjectTypeSecret secret vault object type for HashiCorp vault
+	VaultObjectTypeSecret               string = "secret"
+	defaultVaultAddress                 string = "https://127.0.0.1:8200"
+	defaultKubernetesServiceAccountPath string = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+)
+
+// Provider implements the keyvault-csi-driver provider interface
```
ritazh: nit: `Provider implements the secrets-store-csi-driver provider interface`

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+	ObjectVersion string `json:"objectVersion" yaml:"objectVersion"`
+}
+
+type StringArray struct {
+	Array []string `json:"array" yaml:"array"`
+}
+
+// NewProvider creates a new vault Provider.
+func NewProvider() (*Provider, error) {
+	glog.V(2).Infof("NewProvider")
```
ritazh: nit: `NewProvider for Hashicorp Vault`

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+	p.VaultCACert = attrib["vaultCACertPath"]
+	p.VaultCAPath = attrib["vaultCADirectory"]
+	p.VaultServerName = attrib["vaultTLSServerName"]
+
+	if s := attrib["vaultSkipTLSVerify"]; s != "" {
+		b, err := strconv.ParseBool(s)
+		if err != nil {
+			return err
+		}
+		p.VaultSkipVerify = b
```
ritazh: If `VaultSkipVerify` is not set, let's provide a default value.

pkg/providers/vault/examples/pv-keyvault-csi.yaml
```diff
@@ -0,0 +1,25 @@
...
+    readOnly: true
+    volumeHandle: kv
+    volumeAttributes:
+      providerName: "vault"
+      roleName: "example-role"
+      vaultAddress: "http://10.0.63.109:8200"
+      vaultSkipTLSVerify: "true"
```
ritazh: Does `vaultSkipTLSVerify: "false"` work? If so, how? If not, let's add a //TODO and an issue to track this.

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	return nil
+}
+
+// MountSecretsStoreObjectContent mounts content of the vault object to target path
+func (p *Provider) MountSecretsStoreObjectContent(ctx context.Context, attrib map[string]string, secrets map[string]string, targetPath string, permission os.FileMode) (err error) {
+	roleName := attrib["roleName"]
+	if roleName != "" {
+		p.VaultRole = roleName
+	}
```
ritazh: Do we need to validate `roleName` parameter has some value?

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+	if p.VaultAddress == "" {
+		p.VaultAddress = defaultVaultAddress
+	}
+	glog.V(2).Infof("vault: vault address %s", p.VaultAddress)
+
+	p.VaultCAPem = attrib["vaultCAPem"]
+	p.VaultCACert = attrib["vaultCACertPath"]
+	p.VaultCAPath = attrib["vaultCADirectory"]
+	p.VaultServerName = attrib["vaultTLSServerName"]
+
```
ritazh: Do we need to ensure these parameters are not empty if `VaultSkipVerify == false`

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,370 @@
...
+		b, err := strconv.ParseBool(s)
+		if err != nil {
+			return err
+		}
+		p.VaultSkipVerify = b
+	}
+
+	p.VaultK8SMountPath = attrib["vaultKubernetesMountPath"]
+	if p.VaultK8SMountPath == "" {
+		p.VaultK8SMountPath = "kubernetes"
```
ritazh: Should this be a default const?

ritazh:
Thanks @anubhavmishra 🎉  Left few comments but other than those, LGTM.

pkg/providers/vault/examples/pv-keyvault-csi.yaml
anubhavmishra: yea, that should be part of the documentation of the provider.

pkg/providers/vault/provider.go
anubhavmishra: It should default to false as `p.VaultSkipVerify` is type `bool`.

pkg/providers/vault/provider.go
anubhavmishra: yep, we should ensure it is set.

pkg/providers/vault/provider.go
anubhavmishra: We need one of the values to be set. Otherwise, we default to system certificates. https://github.com/deislabs/secrets-store-csi-driver/pull/23/files/aec594d0f2d07108eff7186238ed117dd43e6b02#diff-e516a1abcc9e4dff2b12ce96f4720d28R204

pkg/providers/vault/provider.go
```diff
@@ -0,0 +1,376 @@
...
+		glog.V(0).Infof("secrets-store csi driver wrote %s at %s", keyValueObject.ObjectPath, targetPath)
+	}
+
+	return nil
+}
+
+// GetKeyValueObjectContent get content of the vault object
+func (p *Provider) GetKeyValueObjectContent(ctx context.Context, objectPath string, objectName string, objectVersion string) (content string, err error) {
+	// Read the jwt token from disk
+	jwt, err := readJWTToken(p.KubernetesServiceAccountPath)
```
seh: Is this using the service account of the pod running the _secrets-store-csi_ program on each node (via a _DaemonSet_)? Ideally we'd use the service account token of the pod requesting the mounting of the secret content, so that we can bind our Vault roles to those service accounts, and not to the service account with which the _secrets-store-csi_ pod runs.

It looks like we wind up here when `(*secretstore.nodeServer).NodePublishVolume` gets invoked. That sounds like it's running in the daemon pod, and not in the context of the mounting pod.

If @anubhavmishra isn't sure, @ritazh may know.

anubhavmishra: Yea, that is correct. Currently, this is using the service account JWT token of the daemon set running the driver. It should be running in the context of the mounting pod. I wasn't quite sure how to best implement it at the time.

seh: Thank you for confirming. Do you have an implementation plan now, or are you open to suggestions? I have an idea, but it comes from ignorance of what the CSI allows these drivers to do.

Should we elevate this to its own issue?

PR #32, opened 2019-05-16
Additional docs to configure a development vault cluster

anubhavmishra:


pkg/providers/vault/README.md
```diff
@@ -7,26 +7,49 @@ Vault and use the Secret Store CSI driver interface to mount them into Kubernete
...
+![Secret Store CSI Driver Vault Provider Demo](./images/secret-store-csi-driver-vault-provider-demo.gif "Secret Store CSI Driver Vault Provider Demo")
 
 ## Prerequisites
 
 The guide assumes the following:
 
-* A Kubernetes cluster up and running.
-* A Vault cluster up and running. Instructions for spinning up a Vault cluster in Kubernetes can be
-found [here](#TODO).
+* A Kubernetes cluster version 1.13.x+ up and running.
```
ritazh: Should we point users to use v1.15.0-alpha2+ for inline volume? to be consistent with https://github.com/deislabs/secrets-store-csi-driver#mount-secret-data-to-resource-through-inline-volume

pkg/providers/vault/README.md
```diff
@@ -7,26 +7,49 @@ Vault and use the Secret Store CSI driver interface to mount them into Kubernete
...
 driver on Kubernetes.
 
+Make sure you have followed the [prerequisites](#prerequisites) specified above before you continue with this guide.
+You should have a development Vault cluster up and running using the [guide](./docs/vault-setup.md) specified above.
+
 ### Install the Secrets Store CSI Driver
 
-Follow instructions [here](../../../README.md#Install-the-Secrets-Store-CSI-Driver) to install the Secret
-Store CSI driver.
+```bash
```
ritazh: If there's nothing specific to Vault here, I recommend that we point users to https://github.com/deislabs/secrets-store-csi-driver#install-the-secrets-store-csi-driver so we have one version of truth going forward.

PR #38, opened 2019-07-01
apiVersion do not match the group and version of the CRD

mskjeret:
There is a mismatch between the deployed CRD (group and version) and the api version used on the CSI.

### DIFF
--- a/charts/secrets-store-csi-driver/templates/csidriver.yaml
+++ b/charts/secrets-store-csi-driver/templates/csidriver.yaml
@@ -1,4 +1,4 @@
-apiVersion: storage.k8s.io/v1beta1
+apiVersion: csi.storage.k8s.io/v1alpha1
 kind: CSIDriver
 metadata:
   name: secrets-store.csi.k8s.com


ritazh:
Thanks @mskjeret! What version of k8s are you using?

`csi.storage.k8s.io/v1alpha1` is no longer supported in v1.14 and `storage.k8s.io/v1beta1` was introduced in v1.14: https://kubernetes-csi.github.io/docs/print.html#breaking-changes

mskjeret:
I see. We are using v1.13. Waiting for AKS to make v1.14 available.

I saw that the CRD, https://github.com/deislabs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver/templates/crd-csi-driver-registry.yaml, still use the old group and v1alpha1 as version.

ritazh:
We are using `csi.storage.k8s.io/v1alpha1` to support older clusters until `csi.storage.k8s.io/v1alpha1` does not work anymore then we will switch to `storage.k8s.io/v1beta1`. I tested `csi.storage.k8s.io/v1alpha1` on a 1.15 cluster and it seems to still work. Let me know if you have any concerns.

mskjeret:
No problem. I was perhaps confused about the order of deployment.

PR #39, opened 2019-07-01
Deploy CRD first

mskjeret:
Make sure the CRD is installed before it is used.


mskjeret:
Hoos for CRD from the helm docs:
crd-install: Adds CRD resources before any other checks are run. This is used only on CRD definitions that are used by other manifests in the chart.

ritazh:
@mskjeret Thank you!

PR #43, opened 2019-07-31
Use go modules

aramase:
Switch to using go modules


aramase:
Closing in favor of https://github.com/deislabs/secrets-store-csi-driver/pull/45

PR #47, opened 2019-08-08
Remove vendor directory

ryysud:
This repo currently uses go modules ( https://github.com/deislabs/secrets-store-csi-driver/pull/45 ) , so I removed  `vendor` directory.


ryysud:
Thank you for your review @aramase.

> The reason we didn't the remove the vendor dir is because the older versions of go can continue to consume it without any change or installing the vendor using go mod vendor.

I understand why the vendor dir hasn't been removed so I close this PR.

PR #54, opened 2019-08-26
Redact access token and client id in logs

aramase:
- Redact access token and client id in Azure provider pod identity scenario
- Reduce log verbosity in deployment
- Update to raw urls

Fixes https://github.com/deislabs/secrets-store-csi-driver/issues/53


aramase:
@ritazh we should be able to do that. I'll make the changes.

aramase:
Created a new PR - https://github.com/deislabs/secrets-store-csi-driver/pull/77. So closing this.

PR #82, opened 2019-09-20
Move providers out of tree

aramase:
Move providers out of tree into their own repos.

- [x] Remove provider code from secrets-store-csi-driver
- [ ] Testing
- [ ] Documentation update


charts/secrets-store-csi-driver/values.yaml
```diff
@@ -8,3 +8,7 @@ logLevel:
 ## Install Default RBAC roles and bindings
 rbac:
   install: true
+
+providers:
```
ritazh: let's make sure at least one provider is specified here, otherwise we will end with `MountVolume.SetUp failed for volume "providers-dir" : hostPath type check failed: /etc/kubernetes/secrets-store-csi-providers is not a directory`

aramase: Instead, should we do `DirectoryOrCreate` in hostPath?

ritazh: we could but the driver should have at least one provider

deploy/secrets-store-csi-driver.yaml
```diff
@@ -42,7 +42,7 @@ spec:
             - name: registration-dir
               mountPath: /registration
         - name: secrets-store
-          image: docker.io/deislabs/secrets-store-csi:v0.0.5
+          image: aramase/secrets-store-csi:v0.0.7
```
ritazh: should we update this image tag after we cut a new release of the csi driver

deploy/provider-vault.yaml
```diff
@@ -0,0 +1,51 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: csi-secrets-store-provider-vault
+    spec:
+      tolerations:
+      containers:
+        - name: provider-vault-installer
+          image: hashicorp/secrets-store-csi-driver-provider-vault:0.0.1
```
ritazh: we could consider making the image tag configurable in values.yaml

deploy/secrets-store-csi-driver.yaml
aramase: Yeah, we should do it after we cut a new release.

charts/secrets-store-csi-driver/values.yaml
aramase: I think we can enable at least one provider in the helm chart once we cut the new release for csi-driver and update the helm chart with the image. In this PR just updating the csi-driver chart is fine with the old image because most of the changes are mounts and hostPath? WDYT?

deploy/secrets-store-csi-driver.yaml
aramase: Done!

deploy/provider-vault.yaml
aramase: Updated the charts to make image tag and repo configurable.

charts/secrets-store-csi-driver/values.yaml
ritazh: SGTM

PR #98, opened 2019-10-24
consolidate controller server and update lock

aramase:
- Moving all controller server methods from `utils` to `controllerServer`.
- Hold lock in `findVolumeInternal` instead of holding in it's caller `findVolume`.

Fixes https://github.com/deislabs/secrets-store-csi-driver/issues/94


pkg/secrets-store/controllerserver.go
```diff
@@ -96,3 +98,49 @@ func (cs *controllerServer) ValidateVolumeCapabilities(ctx context.Context, req
...
+		for _, vol := range cs.vols {
+			if strings.EqualFold(nameOrID, vol.VolumeId) {
+				return vol, true
+			}
+		}
+	}
+	return csi.Volume{}, false
+}
+
+func getProvidersVolumePath() string {
```
aramase: This fn doesn't belong to controller server, but will be removed in a PR to address [issue](https://github.com/deislabs/secrets-store-csi-driver/issues/97).

PR #108, opened 2019-11-01
Update mount tmpfs

### LINKED ISSUE #105: [code review] Update mount logic
Bind mounting "/tmp" to the targetPath seems odd

part of #100

ritazh:
fixes #105


pkg/secrets-store/nodeserver.go
```diff
@@ -160,8 +160,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		parameters["csi.storage.k8s.io/pod.name"] = attrib["csi.storage.k8s.io/pod.name"]
 		parameters["csi.storage.k8s.io/pod.namespace"] = attrib["csi.storage.k8s.io/pod.namespace"]
 	}
-	// to ensure mount bind works, we need to mount before writing content to it
-	err = mounter.Mount("/tmp", targetPath, "", []string{"bind"})
+	// mount before providers can write content to it
+	err = mounter.Mount("tmpfs", targetPath, "tmpfs", []string{})
```
msau42: I believe tmpfs uses up memory on the node. Are we considered about the scalability limits of number secrets that can be mounted on the node?

PR #114, opened 2019-11-02
Update logging 

aramase:
- Switch to `logrus` instead of `glog`
- Print provider stdout logs
- Return provider stderr logs to kubelet

In case of an error -

```
Events:
  Type     Reason       Age   From                                         Message
  ----     ------       ----  ----                                         -------
  Normal   Scheduled    14s   default-scheduler                            Successfully assigned default/nginx-secrets-store-inline-pod-identity to k8s-agentpool1-84472545-vmss000004
  Warning  FailedMount  12s   kubelet, k8s-agentpool1-84472545-vmss000004  MountVolume.SetUp failed for volume "secrets-store-inline" : rpc error: code = Unknown desc = error mounting secret time="2019-11-04T17:18:46Z" level=fatal msg="[error] : failed to get objectType:secret, objectName:secret1, objectVersion:: keyvault.BaseClient#GetSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code=\"Forbidden\" Message=\"Access denied. Caller was not found on any access policy.\\r\\nCaller: appid=xxxxxxxx;oid=xxxxxxxx;numgroups=0;iss=https://sts.windows.net/xxxxxxxxxx/\\r\\nVault: vmss01;location=southcentralus\" InnerError={\"code\":\"AccessDenied\"}"
```

Introduces 3 new flags for logger -
```
--debug=true | false                  sets the log to debug level (Default value: false)
--log-format-json= true | false       sets log formatter to json (Default value: false)
--log-report-caller= true | false     include the calling method as fields in the log (Default value: false)
```


cmd/secrets-store-csi-driver/main.go
```diff
@@ -20,30 +20,30 @@ import (
...
 	if os.Getenv("PROVIDERS_VOLUME_PATH") == "" {
-		glog.Fatalf("providers volume path not provided. Set PROVIDERS_VOLUME_PATH")
+		log.Fatalf("providers volume path not provided. Set PROVIDERS_VOLUME_PATH")
+	}
+
+	log.SetLevel(log.InfoLevel)
+	if *debug {
+		log.SetLevel(log.DebugLevel)
 	}
+	//log.SetFormatter(&log.JSONFormatter{})
```
ritazh: should we expose this as a flag for users to set?

pkg/csi-common/controllerserver-default.go
```diff
@@ -59,7 +59,7 @@ func (cs *DefaultControllerServer) GetCapacity(ctx context.Context, req *csi.Get
 // ControllerGetCapabilities implements the default GRPC callout.
 // Default supports all capabilities
 func (cs *DefaultControllerServer) ControllerGetCapabilities(ctx context.Context, req *csi.ControllerGetCapabilitiesRequest) (*csi.ControllerGetCapabilitiesResponse, error) {
-	glog.V(5).Infof("Using default ControllerGetCapabilities")
+	log.Tracef("Using default ControllerGetCapabilities")
```
ritazh: I don't see where we set the log level to `TraceLevel`

cmd/secrets-store-csi-driver/main.go
aramase: yeah, done!

pkg/csi-common/controllerserver-default.go
aramase: changed this to debug level. We could potentially have the user define the log level using a flag instead of `debug=true| false`. 

It would be `log-level=info|debug|trace`. However I don't think we have that many logs which needs to be split into 3 levels. WDYT?

pkg/csi-common/utils.go
```diff
@@ -127,5 +128,5 @@ func logRedactedRequest(req interface{}) {
 		}
 	}
 	req1.Secrets = redactedSecrets
-	glog.V(5).Infof("GRPC request: %+v", req1)
+	log.Infof("GRPC request: %+v", req1)
```
ritazh: This should probably be debug level or lower.

pkg/secrets-store/nodeserver.go
```diff
@@ -76,21 +77,21 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if !notMnt {
 		// testing original mount point, make sure the mount link is valid
 		if _, err := ioutil.ReadDir(targetPath); err == nil {
-			glog.V(2).Infof("secrets-store - already mounted to target %s", targetPath)
+			log.Debugf("secrets-store - already mounted to target %s", targetPath)
```
ritazh: I think this should be info level

pkg/secrets-store/nodeserver.go
```diff
@@ -242,13 +254,13 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	glog.V(4).Infof("secrets-store: targetPath %s volumeID %s has been unmounted.", targetPath, volumeID)
+	log.Debugf("secrets-store: targetPath %s volumeID %s has been unmounted.", targetPath, volumeID)
```
ritazh: We should probably make this Info level or higher to match all the other mount/unmount logs.

pkg/secrets-store/nodeserver.go
```diff
@@ -242,13 +254,13 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
```
ritazh: Maybe also `log.Errorf` this error.

pkg/secrets-store/secrets-store.go
```diff
@@ -51,13 +52,13 @@ func newControllerServer(d *csicommon.CSIDriver) *controllerServer {
...
-	glog.Infof("Driver: %v ", driverName)
-	glog.Infof("Version: %s", vendorVersion)
+	log.Infof("Driver: %v ", driverName)
+	log.Infof("Version: %s", vendorVersion)
 
 	// Initialize default library driver
 	s.driver = csicommon.NewCSIDriver(driverName, vendorVersion, nodeID)
 	if s.driver == nil {
-		glog.Fatalln("Failed to initialize SecretsStore CSI Driver.")
+		log.Fatalln("Failed to initialize SecretsStore CSI Driver.")
```
ritazh: Let's use `Fatalf` to be consistent.

pkg/secrets-store/secrets-store.go
```diff
@@ -51,13 +52,13 @@ func newControllerServer(d *csicommon.CSIDriver) *controllerServer {
 }
 
 func (s *SecretsStore) Run(driverName, nodeID, endpoint string) {
-	glog.Infof("Driver: %v ", driverName)
-	glog.Infof("Version: %s", vendorVersion)
+	log.Infof("Driver: %v ", driverName)
+	log.Infof("Version: %s", vendorVersion)
```
ritazh: Let's make these two consistent.

pkg/csi-common/controllerserver-default.go
ritazh: IMO, for now, having the log level as `info|debug` should be sufficient. So `debug=true|false` should translate to `debug|info`.  maybe update all the trace level things to debug level? WDYT?

aramase: yeah, that sounds good! Updated all trace to debug.

pkg/secrets-store/secrets-store.go
aramase: done!

pkg/secrets-store/secrets-store.go
aramase: done!

pkg/secrets-store/nodeserver.go
aramase: done!

pkg/secrets-store/nodeserver.go
aramase: done!

pkg/csi-common/utils.go
aramase: done!

pkg/secrets-store/nodeserver.go
aramase: We print out the provider logs, but don't actually have a log which says mount is successful. So kept this debug as other mount logs are also debug.

pkg/secrets-store/nodeserver.go
```diff
@@ -173,60 +174,72 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-			"--permission", string(permissionStr),
-		).Output()
+			"--permission", string(permissionStr)}
+
+		// set log level for provider based on the value set for driver
+		if log.GetLevel() == log.DebugLevel {
+			args = append(args, "--debug", "true")
+		}
+
+		log.Infof("provider command invoked: %s %v", providerBinary, args)
```
ritazh: we need to redact the secrets from the args here

pkg/secrets-store/nodeserver.go
```diff
@@ -173,60 +174,72 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		args := []string{"--attributes", string(parametersStr),
 			"--secrets", string(secretStr),
 			"--targetPath", string(targetPath),
-			"--permission", string(permissionStr),
-		).Output()
+			"--permission", string(permissionStr)}
+
+		// set log level for provider based on the value set for driver
+		if log.GetLevel() == log.DebugLevel {
+			args = append(args, "--debug", "true")
```
ritazh: This needs to be backward compatible with older provider binaries. `unknown flag: --debug`

cmd/secrets-store-csi-driver/main.go
```diff
@@ -20,29 +20,32 @@ import (
...
+		log.Fatalf("providers volume path not provided. Set PROVIDERS_VOLUME_PATH")
+	}
+
+	log.SetLevel(log.InfoLevel)
+	if *debug {
+		log.SetLevel(log.DebugLevel)
+	}
+	if *logFormatJSON {
+		log.SetFormatter(&log.JSONFormatter{})
 	}
```
malnick: While setting up the logger here, you may want to add an option to enable the method logger in logrus: `log.SetReportCaller(true)` - Preferably in a `trace` situation or explicitly since this does add overhead.

pkg/csi-common/driver.go
```diff
@@ -17,13 +17,15 @@ limitations under the License.
...
-	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
+
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 )
 
+// CSIDriver ...
```
malnick: I would complete this comment, maybe `CSIDriver provides a container storage interface driver implementation for the secrets store project`

aramase: thanks, done!

cmd/secrets-store-csi-driver/main.go
aramase: done!

pkg/secrets-store/nodeserver.go
aramase: done!

charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
```diff
@@ -46,7 +46,6 @@ spec:
         - name: secrets-store
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           args:
-            - "--v=3"
```
ritazh: Should we default with `--debug=true` here? also the same change needs to be updated in the yaml in `/deploy`

aramase: A lot of debug logs we had were set at v=5 log level. So I think by default we should leave `debug=false` and enable only if required. 

We should update in `/deploy` too but only when we have a new image for `secrets-store-csi`. B

ritazh: sgtm

PR #117, opened 2019-11-04
validate read-only mode

aramase:
Fixes https://github.com/deislabs/secrets-store-csi-driver/issues/107


pkg/secrets-store/nodeserver.go
```diff
@@ -167,6 +167,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, err
 	}
 	if !isMockProvider(providerName) {
+		// ensure it's read-only
+		if !req.GetReadonly() {
+			return nil, status.Error(codes.InvalidArgument, "read-only mode needs to be true")
```
ritazh: nit: can we update the message to `Readonly is not true in request` to be more consistent with the other status error messages.

aramase: thanks, done!

PR #120, opened 2019-11-05
Validate args before mount

### LINKED ISSUE #118: Validate args before mount

ritazh:
Fixes #118


pkg/secrets-store/nodeserver.go
```diff
@@ -160,12 +160,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		parameters["csi.storage.k8s.io/pod.name"] = attrib["csi.storage.k8s.io/pod.name"]
 		parameters["csi.storage.k8s.io/pod.namespace"] = attrib["csi.storage.k8s.io/pod.namespace"]
 	}
-	// mount before providers can write content to it
```
aramase: We mount this if mock provider/not. So why are we moving it into if/else?

ritazh: This was a bug. When one of the provider parameters is invalid (for example, if provider volume path is invalid), we still mounted the volume since the mount was called before the validation logic. This PR changes that behavior. If any of the provider parameters is invalid, we do not mount and we return an error back to the kubelet.

aramase: ah, got it!

PR #121, opened 2019-11-05
feat: make volume path for provider a flag

### LINKED ISSUE #97: [code review] update provider volume path as a flag
make the provider volume path configurable using a flag instead of the env variable.

part of #92

malnick:
Fixes #97

### DIFF
--- a/cmd/secrets-store-csi-driver/main.go
+++ b/cmd/secrets-store-csi-driver/main.go
@@ -26,20 +26,17 @@ import (
 )
 
 var (
-	endpoint        = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
-	driverName      = flag.String("drivername", "secrets-store.csi.k8s.com", "name of the driver")
-	nodeID          = flag.String("nodeid", "", "node id")
-	debug           = flag.Bool("debug", false, "sets log to debug level")
-	logFormatJSON   = flag.Bool("log-format-json", false, "set log formatter to json")
-	logReportCaller = flag.Bool("log-report-caller", false, "include the calling method as fields in the log")
+	endpoint           = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
+	driverName         = flag.String("drivername", "secrets-store.csi.k8s.com", "name of the driver")
+	nodeID             = flag.String("nodeid", "", "node id")
+	debug              = flag.Bool("debug", false, "sets log to debug level")
+	logFormatJSON      = flag.Bool("log-format-json", false, "set log formatter to json")
+	logReportCaller    = flag.Bool("log-report-caller", false, "include the calling method as fields in the log")
+	providerVolumePath = flag.String("provider-volume", "/etc/kubernetes/secrets-store-csi-providers", "Volume path for provider")
 )
 
 func main() {
 	flag.Parse()
-
-	if os.Getenv("PROVIDERS_VOLUME_PATH") == "" {
-		log.Fatalf("providers volume path not provided. Set PROVIDERS_VOLUME_PATH")
-	}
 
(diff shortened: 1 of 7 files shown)


deploy/secrets-store-csi-driver.yaml
```diff
@@ -47,6 +47,7 @@ spec:
             - "--v=3"
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
+            - "--volume=/etc/kubernetes/secrets-store-csi-providers"
```
ritazh: Can you pls also update this in the helm chart? https://github.com/deislabs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L51-L61

cmd/secrets-store-csi-driver/main.go
```diff
@@ -36,20 +36,16 @@ var (
 	endpoint   = flag.String("endpoint", "unix://tmp/csi.sock", "CSI endpoint")
 	driverName = flag.String("drivername", "secrets-store.csi.k8s.com", "name of the driver")
 	nodeID     = flag.String("nodeid", "", "node id")
+	volumePath = flag.String("volume", "", "Volume path for provider")
```
ritazh: Should we default this flag to `/etc/kubernetes/secrets-store-csi-providers` here?

ritazh:
@malnick Looks like CI failed. Can you pls review the failures?

malnick:
Can do, thanks @ritazh

malnick:
@ritazh Looks like tests are mostly passing now. The `unit test` stage passed before the most recent build but is now failing with ```ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler```. 

If you give me access, I'm happy to manually re-run this stage, I saw a few things about this being related to an issue in golangci but didn't investigate further. It is interesting that it's an intermittent issue.

aramase:
@malnick I re-ran the failing test.

pkg/secrets-store/secrets-store.go
```diff
@@ -18,6 +18,7 @@ package secretsstore
 
 import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/golang/glog"
```
ritazh: Please use logrus instead:

https://github.com/deislabs/secrets-store-csi-driver/blob/master/pkg/secrets-store/secrets-store.go#L24

deploy/secrets-store-csi-driver.yaml
ritazh: hmm can we hold off on changes in this yaml until after we cut a new release or the deployment will fail since this flag is new and the previous release requires the env var.

cmd/secrets-store-csi-driver/main.go
```diff
@@ -32,6 +32,7 @@ var (
 	debug           = flag.Bool("debug", false, "sets log to debug level")
 	logFormatJSON   = flag.Bool("log-format-json", false, "set log formatter to json")
 	logReportCaller = flag.Bool("log-report-caller", false, "include the calling method as fields in the log")
+	volumePath      = flag.String("volume", "/etc/kubernetes/secrets-store-csi-providers", "Volume path for provider")
```
aramase: nit: I think we should call it `provider-volume` just to be more specific.

deploy/secrets-store-csi-driver.yaml
ritazh: @malnick can you pls take a look at ^

malnick:
@ritazh I removed the changes to the deploy/csi...yaml file. Thanks!

Issue #106, opened 2019-11-01
[code review] idempotency of NodePublish/Unpublish


ritazh:
* Double check idempotency of NodePublish/Unpublish. For example, if NodePublish gets called twice, and it's already correctly mounted, we should return success. Looking at the current code, it seems like we'll bind mount it twice? Which can potentially cause mounts to leak on cleanup.  Similarly, Unpublish should return success if it's already unpublished.
* Usage of IsLikelyNotMountPoint() also doesn't work if the bind mount is from the same fs as the kubelet's rootfs (if you look at the implementation, it just checks if the parent directory is in the same fs or not). 
* results from https://github.com/deislabs/secrets-store-csi-driver/blob/51d64fcf33d62b84143f666ed8bc387011652de5/pkg/secrets-store/nodeserver.go#L220 are unused?

part of #100

aramase:
Closed with https://github.com/deislabs/secrets-store-csi-driver/pull/124

PR #124, opened 2019-11-08
check is mounted before unmount

### LINKED ISSUE #123: Check is mounted before calling unmount
```
level=error msg="GRPC error: rpc error: code = Internal desc = Unmount failed: exit status 32\nUnmounting arguments: /var/lib/kubelet/pods/xxxx/volumes/kubernetes.io~csi/secrets-store-inline/mount\nOutput: umount: /var/lib/kubelet/pods/xxxx/volumes/kubernetes.io~csi/secrets-store-inline/mount: not mounted\n\n"
```

aramase:
- Check if mounted before unmount 
- Use results of `IsLikelyNotMount` after node publish volume is complete.

Fixes #123, #106

### DIFF
--- a/pkg/secrets-store/nodeserver.go
+++ b/pkg/secrets-store/nodeserver.go
@@ -249,7 +249,6 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if notMnt {
 		return nil, fmt.Errorf("after MountSecretsStoreObjectContent, notMnt: %v", notMnt)
 	}
-	log.Debugf("after MountSecretsStoreObjectContent, notMnt: %v", notMnt)
 	return &csi.NodePublishVolumeResponse{}, nil
 }
 
@@ -267,6 +266,7 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 
 	// check if target path is mount point
 	notMnt, err := mount.New("").IsLikelyNotMountPoint(targetPath)
+	// if not mount point then no need to unmount
 	if (err != nil && os.IsNotExist(err)) || notMnt {
 		log.Infof("target path %s is not likely a mount point, notMnt %v", targetPath, notMnt)
 		return &csi.NodeUnpublishVolumeResponse{}, nil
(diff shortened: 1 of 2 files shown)


pkg/secrets-store/nodeserver.go
```diff
@@ -259,14 +264,30 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	if (err != nil && os.IsNotExist(err)) || notMnt {
+		log.Infof("target path %s is not likely a mount point, notMnt %v", targetPath, notMnt)
+		return &csi.NodeUnpublishVolumeResponse{}, nil
+	}
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+	// if mount point, ensure mount link is still valid
+	if !notMnt {
+		if _, err = ioutil.ReadDir(targetPath); err != nil {
```
ritazh: If the target is mounted, when would it fail on read?

pkg/secrets-store/nodeserver.go
```diff
@@ -240,9 +241,13 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	if err != nil {
-		log.Errorf("Error checking IsLikelyNotMountPoint: %v", err)
+		log.Errorf("Error checking targetPath %s IsLikelyNotMountPoint: %v", targetPath, err)
 		return nil, err
 	}
+	// this means the mount didn't succeed
+	if notMnt {
+		return nil, fmt.Errorf("after MountSecretsStoreObjectContent, notMnt: %v", notMnt)
+	}
 	log.Debugf("after MountSecretsStoreObjectContent, notMnt: %v", notMnt)
```
ritazh: since we are returning the error if not mounted, we can remove this log

aramase: Done!

pkg/secrets-store/nodeserver.go
aramase: As long it's a valid mnt point, we should unmount it. So removed this check.

PR #135, opened 2019-12-02
Add driver-provider version check

aramase:
fixes https://github.com/deislabs/secrets-store-csi-driver/issues/130


aramase:
cc @malnick @kalafut

pkg/version/version.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if minProviderVersions == "" {
+		return providerVersionMap
+	}
+
+	providers := strings.Split(minProviderVersions, ",")
+	for _, provider := range providers {
+		provider = strings.TrimSpace(provider)
+
+		pv := strings.Split(provider, "=")
```
ritazh: missing error handling around existence of "="

pkg/version/version.go
```diff
@@ -0,0 +1,98 @@
...
+	"strings"
+
+	"github.com/blang/semver"
+	log "github.com/sirupsen/logrus"
+)
+
+// providerVersion holds current provider version
+type providerVersion struct {
+	Version string `json:"version"`
+	Date    string `json:"date"`
```
ritazh: What is `Date`? how will it be used and who will use it?

pkg/version/version.go
```diff
@@ -0,0 +1,98 @@
...
+// GetMinimumProviderVersions creates a map with provider name and minimum version
+// supported with this driver.
+func GetMinimumProviderVersions(minProviderVersions string) map[string]string {
+	providerVersionMap := make(map[string]string)
+
+	if minProviderVersions == "" {
+		return providerVersionMap
+	}
+
+	providers := strings.Split(minProviderVersions, ",")
```
ritazh: error handling for ","

pkg/version/version_test.go
```diff
@@ -0,0 +1,111 @@
...
+		{
+			desc:                "white space in min provider versions",
+			minProviderVersions: "provider1=0.0.2, provider2=0.0.4",
+			expectedMap:         map[string]string{"provider1": "0.0.2", "provider2": "0.0.4"},
+		},
+		{
+			desc:                "more white space in min provider versions",
+			minProviderVersions: "provider1=0.0.2 , provider2=0.0.4",
+			expectedMap:         map[string]string{"provider1": "0.0.2", "provider2": "0.0.4"},
+		},
```
ritazh: maybe also add few more tests for invalid semver and test for ',' and '='

pkg/version/version.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/blang/semver"
+	log "github.com/sirupsen/logrus"
+)
+
+// providerVersion holds current provider version
+type providerVersion struct {
+	Version string `json:"version"`
+	Date    string `json:"date"`
+	// MinDriverVersion is minimum driver version the provider works with
+	MinDriverVersion string `json:"minDriverVersion"`
```
ritazh: maybe make a note about how this value will be used in the future?

pkg/secrets-store/nodeserver.go
```diff
@@ -203,6 +205,18 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 		log.Infof("Calling provider: %s", providerName)
 		providerBinary := fmt.Sprintf("%s/%s/provider-%s", providerVolumePath, providerName, providerName)
+
+		// check if provider is compatible with driver
+		if v, exists := ns.minProviderVersions[providerName]; exists {
```
ritazh: return an error when providerName does not exist?

aramase: I was thinking if the compatible version isn't explicitly defined then we can assume all provider versions are compatible with the driver. But after a little more thought, I think we should explicitly define the versions in the deployment and error out here to avoid scenarios where deployment manifest is directly updated with the new image instead of applying the whole new manifest.

pkg/version/version.go
aramase: Will add error handling for "=". That along with the base case check before this should ensure if we catch any error in format.

pkg/version/version.go
aramase: `Date` here is the build date and time for the provider image. This can be used in debug logs to see the build date is latest if the same image tag is rebuilt in some corner scenario.

pkg/version/version_test.go
aramase: done!

pkg/version/version.go
aramase: done!

pkg/version/version.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// splitting on , delimiter will result in array of provider=value string
+	providers := strings.Split(minProviderVersions, ",")
+	for _, provider := range providers {
+		provider = strings.TrimSpace(provider)
+
+		pv := strings.Split(provider, "=")
+		// validate its expected format provider=version
+		if len(pv) != 2 || len(pv[0]) == 0 || len(pv[1]) == 0 {
```
kalafut: Probably want to TrimSpace the `pv` parts before the length tests, otherwise something like `       =      ` would be valid.

pkg/secrets-store/secrets-store.go
```diff
@@ -38,11 +39,17 @@ func GetDriver() *SecretsStore {
 	return &SecretsStore{}
 }
 
-func newNodeServer(d *csicommon.CSIDriver, providerVolumePath string) *nodeServer {
-	return &nodeServer{
-		DefaultNodeServer:  csicommon.NewDefaultNodeServer(d),
-		providerVolumePath: providerVolumePath,
+func newNodeServer(d *csicommon.CSIDriver, providerVolumePath, minProviderVersions string) (*nodeServer, error) {
+	// get a map of provider and compatible version
+	minProviderVersionsMap, err := version.GetMinimumProviderVersions(minProviderVersions)
```
ritazh: should check for len == 0

pkg/secrets-store/nodeserver.go
ritazh: I mean I did not see any handling of if  the providername does not exist in the map.

pkg/version/version.go
```diff
@@ -0,0 +1,106 @@
...
+			return providerVersionMap, fmt.Errorf("min provider version not defined in expected format, got %+v", pv)
+		}
+
```
kalafut: A couple other checks for consideration:

- Is the value (pv[1]) valid semver? This would be caught later, but it might be a better spot to error here.
- Are duplicate provider names an error?  `a=5,b=9,a=10` seems like it should be an error.

pkg/secrets-store/nodeserver.go
aramase: @ritazh I haven't pushed all the changes yet. Once I push that change CI will fail until we update the helm chart with the new min provider version flag.

pkg/version/version.go
aramase: thank you @kalafut! I've added the changes. PTAL!

pkg/secrets-store/secrets-store.go
aramase: done!

pkg/secrets-store/nodeserver.go
aramase: done!

pkg/version/version.go
aramase: done!

pkg/version/version.go
aramase: done!

pkg/version/version.go
kalafut: LGTM!

aramase:
I've opened a [PR](https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/16) in azure provider to support `--version` flag. We need a similar change in the vault provider too.

cc @ritazh @kalafut

ritazh:
I think the azure.bat needs to be updated now

PR #142, opened 2019-12-17
Add liveness probe

aramase:
fixes https://github.com/deislabs/secrets-store-csi-driver/issues/119

### DIFF
--- a/pkg/secrets-store/identityserver.go
+++ b/pkg/secrets-store/identityserver.go
@@ -17,9 +17,9 @@ limitations under the License.
 package secretsstore
 
 import (
-	"context"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	csicommon "github.com/deislabs/secrets-store-csi-driver/pkg/csi-common"
+	"golang.org/x/net/context"
 
 	wrappers "github.com/golang/protobuf/ptypes/wrappers"
 )
--- a/pkg/secrets-store/identityserver.go
+++ b/pkg/secrets-store/identityserver.go
@@ -1,0 +1,37 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


pkg/secrets-store/identityserver.go
```diff
@@ -0,0 +1,37 @@
...
+type identityServer struct {
+	*csicommon.DefaultIdentityServer
+}
+
+// Probe check whether the plugin is running or not.
+// This method does not need to return anything.
+// Currently the spec does not dictate what you should return either.
+// Hence, return an empty response
+func (ids *identityServer) Probe(ctx context.Context, req *csi.ProbeRequest) (*csi.ProbeResponse, error) {
+	return &csi.ProbeResponse{Ready: &wrappers.BoolValue{Value: true}}, nil
```
ritazh: The comment and the implementation are different. This is always returning ready. Is that right?

pkg/secrets-store/secrets-store.go
```diff
@@ -27,16 +27,20 @@ import (
...
+// NodeServer CSI interfaces.
 type SecretsStore struct {
 	driver *csicommon.CSIDriver
 	ns     *nodeServer
 	cs     *controllerServer
+	ids    *identityServer
 }
 
 var (
 	vendorVersion = "0.0.3"
```
ritazh: Just realized this should have been updated for each release.

pkg/secrets-store/identityserver.go
aramase: I'll update the comment. The `liveness-probe` sidecar container uses the Probe RPC call to check if ready is true. Currently we don't need any other criteria other than successfully able to make Probe RPC call for liveness probe. That's the reason it always returns true for ready.

aramase: Updated the comment.

pkg/secrets-store/secrets-store.go
aramase: done! updated to reflect latest version. Going forward we can bump this along with other places before cutting new release.

PR #146, opened 2020-01-10
Remove providers from chart

ritazh:
- Removed providers from chart
- Removed providers from deploy/
- Updated minimum compatible provider version check optional
- Update e2e to use deployment yaml from provider repo


test/bats/azure.bats
```diff
@@ -23,8 +24,19 @@ setup() {
...
           --set image.pullPolicy="IfNotPresent" \
           --set image.repository="e2e/secrets-store-csi" \
-          --set image.tag=$IMAGE_TAG \
-          --set providers.azure.enabled=true
+          --set image.tag=$IMAGE_TAG
+  assert_success
+}
+
+@test "install azure provider" {
+  run kubectl apply -f $PROVIDER_YAML
```
aramase: this will deploy the provider manifest in the default namespace. This will not be an issue - but maybe we should deploy all the components in the same namespace or maybe add a note here just so we are aware of it.

ritazh: Added $NAMESPACE to doc and e2e for vault and azure.

charts/secrets-store-csi-driver/values.yaml
```diff
@@ -13,16 +13,5 @@ livenessProbe:
...
-    tag: 0.0.3
-    imagePullPolicy: Always
-
-  vault:
-    enabled: false
-    repository: hashicorp/secrets-store-csi-driver-provider-vault
-    tag: 0.0.4
-    imagePullPolicy: Always
+## Minimum Provider Versions (optional)
+minimumProviderVersions:
```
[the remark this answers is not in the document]
ritazh: Done

PR #160, opened 2020-02-27
update repo path to sigs.k8s.io

aramase:
Since the repo move to `kubernetes-sigs` is complete, we can update the path alias to `sigs.k8s.io`.


aramase:
/assign @ritazh 

The test-infra changes have been merged and it looks good. Need to fix the `e2e-vault` job in prow. I'll take care of that next. Will also remove the travis job in a separate PR (until then we have tests with prow and travis :smile: )

Issue #115, opened 2019-11-04
Windows support


ritazh:

aramase:
#164 adds windows support for drivers
https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/28 will add windows support for azure provider
  - pod identity mode will not be supported

Will close this issue when official windows driver and provider images are available

aramase:
Windows support has been added for the driver and azure provider. 

Pending: doc update with instructions for windows.

PR #164, opened 2020-03-04
add support for windows

aramase:
fixes https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/115


deploy/secrets-store-csi-driver-windows.yaml
```diff
@@ -0,0 +1,78 @@
...
+        beta.kubernetes.io/os: windows
+      serviceAccountName: secrets-store-csi-driver
+      containers:
+        - name: node-driver-registrar
+          image: mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v1.3.0-pipeline
+          args:
+            - --v=5
+            - "--csi-address=unix://C:\\csi\\csi.sock"
+            - "--kubelet-registration-path=C:\\var\\lib\\kubelet\\plugins\\csi-secrets-store\\csi.sock"
+          env:
```
ritazh: Is there an equivalent command we can run to remove the socket for preStop?
```
rm -rf /registration/secrets-store.csi.k8s.io-reg.sock
```

deploy/secrets-store-csi-driver-windows.yaml
```diff
@@ -0,0 +1,78 @@
...
+    metadata:
+      labels:
+        app: csi-secrets-store
+    spec:
+      nodeSelector:
+        beta.kubernetes.io/os: windows
+      serviceAccountName: secrets-store-csi-driver
+      containers:
+        - name: node-driver-registrar
+          image: mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v1.3.0-pipeline
```
ritazh: What is the equivalent of this image compare to `quay.io/k8scsi/csi-node-driver-registrar`? The latest at https://github.com/kubernetes-csi/node-driver-registrar/releases is `1.2.0`

Do we also need to update this tag for the linux deployment?

pkg/secrets-store/nodeserver.go
```diff
@@ -282,22 +268,26 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+			err = os.RemoveAll(file)
+			if err != nil {
+				log.Errorf("failed to remove file %s, err: %v", file, err)
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		}
 	}
-	// Unmounting the target
-	err = mount.New("").Unmount(targetPath)
+	err := mount.CleanupMountPoint(targetPath, ns.mounter, false)
```
ritazh: Can you double check that `extensiveMountPointCheck` should be set to `false` for Windows, which uses `IsLikelyNotMountPoint`
https://github.com/kubernetes/utils/blob/master/mount/mount_helper_common.go#L26-L28

ritazh:
Left few comments but mostly looks great! Would be great to think about how to add e2e tests for Windows as part of CI

Makefile
```diff
@@ -39,8 +39,12 @@ sanity-test:
 	go test -v ./test/sanity
 build: setup
 	CGO_ENABLED=0 GOOS=linux go build -a -ldflags ${LDFLAGS} -o _output/secrets-store-csi ./cmd/secrets-store-csi-driver
+build-windows: setup
+	CGO_ENABLED=0 GOOS=windows go build -a -ldflags ${LDFLAGS} -o _output/secrets-store-csi.exe ./cmd/secrets-store-csi-driver
 image: build
 	docker build --no-cache -t $(IMAGE_TAG) -f Dockerfile .
+image-windows: build-windows
+	docker build --no-cache -t $(IMAGE_TAG) -f windows.Dockerfile .
```
ritazh: nit: not sure what is the best naming convention for Dockerfile. here are few from other projects to consider:
https://github.com/moby/moby/blob/master/Dockerfile.windows
https://github.com/kubernetes/kubernetes/blob/master/test/images/busybox/Dockerfile_windows

WDYT?

pkg/secrets-store/nodeserver.go
aramase: I've tested with `extensiveMountPointCheck` true too, but the List mountpoints interface is not implemented for windows which means it always returns not mount. I'll double check again.

deploy/secrets-store-csi-driver-windows.yaml
aramase: > What is the equivalent of this image compare to quay.io/k8scsi/csi-node-driver-registrar? The latest at https://github.com/kubernetes-csi/node-driver-registrar/releases is 1.2.0

There is still no official build for windows yet. We need the build from master as that contains all the changes required to make it work on windows. `v1.3.0-pipeline` is building the master.

> Do we also need to update this tag for the linux deployment?

Yes, we should update to the latest available release. I'll open a PR to bump it up.

deploy/secrets-store-csi-driver-windows.yaml
aramase: Will try it out and see if we can use a windows command to remove it

pkg/secrets-store/nodeserver.go
aramase: Validated this! `IsMountPoint` done for extensive check calls list mount interface (which isn't implemented for windows). So it always returns not mnt.

deploy/secrets-store-csi-driver-windows.yaml
aramase: Added the prestop command to delete the socket

Makefile
aramase: Having the `*.Dockerfile` file naming convention makes it possible for editors to recognize this as a docker file and provides syntax highlighting. That was one of the reasons for naming it this way.

ritazh: ok that makes sense!

ritazh:
lgtm

PR #178, opened 2020-03-24
update manifests and release for v0.0.9

aramase:


test/bats/vault.bats
```diff
@@ -10,7 +10,7 @@ NAMESPACE=dev
 PROVIDER_YAML=https://raw.githubusercontent.com/hashicorp/secrets-store-csi-driver-provider-vault/master/deployment/provider-vault-installer.yaml
 
 @test "install helm chart with e2e image" {
-  run helm install charts/secrets-store-csi-driver -n csi-secrets-store --namespace $NAMESPACE \
+  run kubectl create ns $NAMESPACE && helm install charts/secrets-store-csi-driver -n csi-secrets-store --namespace $NAMESPACE \
```
aramase: With the latest helm release `v2.16.4`, helm no longer creates the namespace if not exists. So the user needs to ensure the namespace exists before running helm install with `--namespace`

aramase:
/hold

seems to be some issue with latest helm

Issue #179, opened 2020-03-24
Add e2e test suite


aramase:
- Support testing on windows

/assign

ritazh:
- [ ] Support testing cluster upgrade
- [ ] Support testing cluster scaling
- [ ] helm upgrade

PR #181, opened 2020-04-06
update tests for windows

aramase:
- Updates makefile to add rule for windows-e2e
- Adds helm chart for windows

fixes https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/179

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,13 +125,14 @@ ifdef TEST_WINDOWS
 			--set windows.image.pullPolicy="IfNotPresent" \
 			--set windows.image.repository=$(REGISTRY)/$(IMAGE_NAME) \
 			--set windows.image.tag=$(IMAGE_VERSION) \
-			--set windows.enabled=true
+			--set windows.enabled=true \
+			--set linux.enabled=false
 else
 		helm install csi-secrets-store charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
-			--set image.pullPolicy="IfNotPresent" \
-			--set image.repository="e2e/secrets-store-csi" \
-			--set image.tag=$(IMAGE_VERSION) \
-			--set image.pullPolicy="IfNotPresent"
+			--set linux.image.pullPolicy="IfNotPresent" \
+			--set linux.image.repository="e2e/secrets-store-csi" \
+			--set linux.image.tag=$(IMAGE_VERSION) \
+			--set linux.image.pullPolicy="IfNotPresent"
 endif
 
 .PHONY: e2e-azure
--- a/Makefile
+++ b/Makefile
@@ -95,14 +95,14 @@ setup-kind:
 	# Create kind cluster
 	kind create cluster --config kind-config.yaml --image kindest/node:v${KUBERNETES_VERSION}
 	# Build image
-	REGISTRY="e2e" IMAGE_VERSION=${E2E_IMAGE_VERSION} make image
+	REGISTRY="e2e" make image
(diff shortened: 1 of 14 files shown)


aramase:
/hold

(using prow to test the PR as having trouble with dev machine). Will remove hold and assign when ready for review

charts/secrets-store-csi-driver/templates/csidriver.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
 spec:
   podInfoOnMount: true
   attachRequired: false
-{{- if and ( ge (.Capabilities.KubeVersion.Major | int) 1 ) ( ge (.Capabilities.KubeVersion.Minor | int) 16 ) }}
-  volumeLifecycleModes: # Added in Kubernetes 1.16 with default mode of Persistent. Secrets store csi driver needs Ephermeral to be set.
+{{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion }}
```
aramase, line 8: Using int compare does not work for pre-release versions (alpha/beta). So switched this to use `semverCompare` instead.

charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.windows.enabled false }}
```
ritazh, line 1: What about hybrid clusters?

Makefile
```diff
@@ -65,37 +68,76 @@ endif
...
+	curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
+
+.PHONY: e2e-teardown
+e2e-teardown:
+	helm delete csi-secrets-store --namespace default
+
+.PHONY: install-driver
+install-driver:
+ifdef TEST_WINDOWS
+		helm install csi-secrets-store charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
```
ritazh, line 124: maybe use the same helm install cmd but set diff variables in the if/else?

charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
aramase: Updated to support hybrid scenarios

Makefile
aramase: The variables for linux and windows are different which is why the helm install cmd is different in if/else

aramase:
@ritazh PTAL when you get a chance! Updated the charts to support hybrid scenarios and also added liveness probe sidecar for windows.

charts/secrets-store-csi-driver/values.yaml
```diff
@@ -1,7 +1,16 @@
...
+    repository: docker.io/deislabs/secrets-store-csi
+    tag: v0.0.9
+    pullPolicy: Always
+
+windows:
+  enabled: false
+  image:
+    repository: mcr.microsoft.com/k8s/csi/secrets-store/driver
+    tag: v0.0.9
+    pullPolicy: IfNotPresent
```
ritazh, line 13: Why is this `IfNotPresent`

aramase: In general, windows images could be large and we want to prevent pulling image `Always`. wdyt?

ritazh: yea that makes sense

PR #189, opened 2020-04-22
Adds image automated build

claudiubelu:
In order for the secret store CSI image to be automatically built and published to the staging registry (from which it will be promoted), the ``cloudbuild.yaml`` file has been added.

The file was added in conformance with [1].

Adds the docker folder, which contains several items:
- cloudbuild.yaml
- BASEIMAGE: the base image to use when building an image for a certain os/arch
  (or os/arch/version for Windows).
- BASEIMAGE_CORE: for Windows images, from which image to copy necessary files (DLLs).
- Makefile
- build.sh: script that can build, push, and create the manifest list.
- Dockerfile: needed to build the Linux docker image. Can now accept the a
  BASEIMAGE arg, which can be useful when building multi-arch images.
- windows.Dockerfile: needed to build the Windows docker image. Can now accept
  BASEIMAGE and BASEIMAGE_CORE args, which can be useful when building images for multiple
  Windows versions.

The image building process will be triggered when changes to the files in the docker changes
are made (for example, you bump the image version, so a new image is built).

[1] https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

Adds support for multiple Windows versions.

The image will be created with the name ``gcr.io/k8s-staging-csi-secrets-store/driver:VERSION``.

Bumps the driver's image version.

Related PR: https://github.com/kubernetes/test-infra/pull/17335

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,9 @@ build: setup
 build-windows: setup
 	CGO_ENABLED=0 GOOS=windows go build -a -ldflags ${LDFLAGS} -o _output/secrets-store-csi.exe ./cmd/secrets-store-csi-driver
 image:
-	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f Dockerfile --platform="linux/amd64" --output "type=docker,push=false" .
+	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f docker/Dockerfile --platform="linux/amd64" --output "type=docker,push=false" .
 image-windows:
-	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f windows.Dockerfile --platform="windows/amd64" --output "type=docker,push=false" .
+	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f docker/windows.Dockerfile --platform="windows/amd64" --output "type=docker,push=false" .
 clean:
 	-rm -rf _output
 setup: clean
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,3 +1,5 @@
+ARG BASEIMAGE=us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:v2.1.0
+
 FROM golang:1.13.10-alpine3.10 as builder
 WORKDIR /go/src/sigs.k8s.io/secrets-store-csi-driver
 ADD . .
@@ -6,7 +8,7 @@ ARG TARGETOS
 ARG LDFLAGS
 RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -ldflags "${LDFLAGS}" -o _output/secrets-store-csi ./cmd/secrets-store-csi-driver
 
-FROM us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:v2.1.0
+FROM $BASEIMAGE
 COPY --from=builder /go/src/sigs.k8s.io/secrets-store-csi-driver/_output/secrets-store-csi /secrets-store-csi
 RUN clean-install ca-certificates cifs-utils mount
 
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,0 +1,30 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REGISTRY?=docker.io/deislabs
+IMAGE_NAME=driver
+IMAGE_VERSION?=v0.0.11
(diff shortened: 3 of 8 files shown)


aramase:
@claudiubelu The e2e tests are failing because they're looking for Dockerfile in the repo root. Once you update that to `docker/Dockerfile` in the Makefile - https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/Makefile#L53-L56 and https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/Makefile#L103-L104 the tests should pass.

claudiubelu:
I've tried building this image with the added script. The image I've built is at ``claudiubelu/driver:v0.0.10`` (it's a manifest list with all the os/arch/versions described in the BASEIMAGE file).

claudiubelu:
FWIW, I've updated the way the image is built, so the Windows nodes are not required at all: https://paste.ubuntu.com/p/phqKvJDDbt/

From what I've tried, containers using the image ``claudiubelu/driver:v0.0.10`` have started on Windows Server 1809, 1903, and 1909. If someone could also check them, that would also be great.

The more difficult part of this was actually tagging the manifest list with the proper ``os.version`` platform entries for the Windows images, which are mandatory for Windows. Without those, a Windows node will try to pull the first image that fits the platform and arch from the manifest list, even if it won't match the OS version, and thus fail to start the container. ``docker manifest annotate`` doesn't allow you to set the ``os.version``, at least, in the current version, so I had to use ``manifest-tool`` to create and push the manifest list, which supports adding the ``os.version``.

I've also updated the image in ``cloudbuild.yaml`` to a newer one that also supports buildx.

aramase:
@claudiubelu We recently merged the PR to use `buildx` for building multi-arch images - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/205. 

> so I had to use manifest-tool to create and push the manifest list, which supports adding the os.version.

This is great!

Makefile
```diff
@@ -51,9 +52,15 @@ build: setup
 build-windows: setup
 	CGO_ENABLED=0 GOOS=windows go build -a -ldflags ${LDFLAGS} -o _output/secrets-store-csi.exe ./cmd/secrets-store-csi-driver
 image:
-	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f Dockerfile --platform="linux/amd64" --output "type=docker,push=false" .
+	docker buildx build --no-cache --build-arg LDFLAGS=${LDFLAGS} -t $(IMAGE_TAG) -f docker/Dockerfile --platform="linux/amd64" --output "type=docker,push=false" .
```
aramase, line 55: there is no `docker/` dir?

claudiubelu:
> 
> 
> @claudiubelu We recently merged the PR to use `buildx` for building multi-arch images - #205.
> 
> > so I had to use manifest-tool to create and push the manifest list, which supports adding the os.version.
> 
> This is great!

Hmm, from what I see, there is still only linux/amd64 and windows/amd64.  There is no arm, amr64, etc. archs added.

Makefile
claudiubelu: Indeed, I moved some things on rebase (there was a merge conflict). Because of some recent changes, I had to rewrite some bits of the building process, and I moved everything in the root directory.

I still prefer having the docker folder. The reason is because we're configuring the Image Promoter to build the image when changes to files in the ``docker/.*`` are being made (for example, you'd trigger the builder on an image version bump). Having everything on the root directory would mean that the builder would be triggered on every PR, which is unnecessary.

mkumatag: @claudiubelu I'm just wondering if we can make it more architecture natural so that we can build other platforms as well like arm, ppc64le and s390x as well?

claudiubelu: I did, in the ``build.sh`` below.

As I mentioned before, I've made a separate ``docker/Makefile`` which is going to be used by the Image Builder / Promoter. This target is still being used for testing, and is called in this ``Makefile``. This only builds the ``linux/amd64`` image, which is sufficient for testing, while mine builds all images, which takes a lot more time, especially due to the Windows images.

mkumatag: cool

docker/Dockerfile
```diff
@@ -6,7 +8,7 @@ ARG TARGETOS
 ARG LDFLAGS
 RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -ldflags "${LDFLAGS}" -o _output/secrets-store-csi ./cmd/secrets-store-csi-driver
```
jingxu97, line 9: If the Dockerfile container RUN, then I think we cannot use buildx to build windows container image on a linux machine. The workaround is to remote go build outside as a separate step.

jingxu97: oh, I see this is only for Linux. just ignore my comment.

docker/windows.Dockerfile
```diff
@@ -8,7 +11,8 @@ RUN CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -ldflags "${LDFL
```
jingxu97, line 11: If the Dockerfile container RUN, then I think we cannot use buildx to build windows container image on a linux machine. The workaround is to remote go build outside as a separate step.

claudiubelu: Indeed, but the RUN command is being run on a Linux image (``golang:1.13.10-alpine3.10``), which is why the actual build itself succeeded. I have posted the logs from my build.

docker/Makefile
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REGISTRY?=docker.io/deislabs
+IMAGE_NAME=driver
+IMAGE_VERSION?=v0.0.10
```
aramase, line 17: Could you update this to `v0.0.11` as that's the last release we cut?

claudiubelu:
Updated version

PR #220, opened 2020-06-03
set gcr_cloud_build for buildx

aramase:

Fixes #


claudiubelu:
I wonder if it's ok to call ``gcr_cloud_build`` twice. And putting it in the build script would mean that people trying to build and push that image to a different registry than gcr.io would be inconvenienced. I wonder if we can't put it in the ``cloudbuild.yaml``, like here: https://github.com/kubernetes/kubernetes/blob/master/build/pause/cloudbuild.yaml

aramase:
@claudiubelu Makes sense! Let me make that change.

aramase:
@claudiubelu Updated, PTAL!

PR #222, opened 2020-06-04
use docker for generating manifests

### LINKED ISSUE #221: Switch to using docker manifest for image push
Switch to using docker manifest for image push as the current image push job fails due to auth with manifest tool. 

```
time="2020-06-03T22:06:23Z" level=fatal msg="V2 registry PUT of manifest list failed: Put https://gcr.io/v2/k8s-staging-csi-secrets-store/driver/manifests/v0.0.11: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication"
```

We can generate the docker manifest and push using `docker manifest`

**Environment:**

aramase:

Using `docker manifest` to generate image manifests and push. Tested with image `aramase/secrets-store-csi:buildx00`.

Fixes #221 

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -99,49 +99,16 @@ build_and_push() {
   done
 }
 
-ensure_manifest_tool() {
-  if ! [[ -x "$(command -v manifest-tool)" ]]; then
-    wget "https://github.com/estesp/manifest-tool/releases/download/v1.0.2/manifest-tool-linux-amd64" -O /usr/bin/manifest-tool
-    chmod +x /usr/bin/manifest-tool
-  fi
-}
-
 # This function will create and push the manifest list for the image
 manifest() {
-  ensure_manifest_tool
-
-  echo "Building manifest list .yaml file for ${IMAGE_TAG}"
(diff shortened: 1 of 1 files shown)


claudiubelu:
``docker manifest create`` was the first thing I've tried too. I really wished it was ok, it would have been simpler. :)

The problem is that ``docker buildx`` doesn't set the ``os.version`` field for the Windows images that it builds, and then ``docker manifest create`` doesn't set those fields either, since it doesn't have any. Plus, ``docker manifest annotate`` doesn't allow you to set the ``os.version`` either, since it's using the manifest list v1 format for some reason.

That is why I was using ``manifest-tool``. It allowed me to set that ``os.version``. Without that ``os.version``, your manifest list can only support one version of Windows, since docker will pull the first one that matches the OS it's running on: windows/amd64. When there are multiple Windows images, it also looks at the ``os.version`` field to match the host OS version as closely as possible.

aramase:
/close

The changes have been included in https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/225

PR #238, opened 2020-06-26
set context for provider binary calls

### LINKED ISSUE #236: [CreateContainerConfigError secret not found] sync secrets option
secret class
```
apiVersion: secrets-store.csi.x-k8s.io/v1alpha1
kind: SecretProviderClass
metadata:
  name: azure-kv-test-env-vars
  namespace: kube-infrastructure
spec:
  secretObjects:                              # [OPTIONAL] SecretObject defines the desired state of synced K8s secret objects
  - data:
    - key: password                           # data field to populate
      objectName: $SECRET_NAME                # name of the mounted content to sync. this could be the object name or the object alias
    secretName: $SECRET_NAME                  # name of the Kubernetes Secret object
    type: Opaque                              # type of the Kubernetes Secret object e.g. Opaque, kubernetes.io/tls
  provider: azure
  parameters:
    usePodIdentity: "true"             # [OPTIONAL] if not provided, will default to "false"
    keyvaultName:  $KEYVAULT_NAME      # the name of the KeyVault
    objects:  |
      array:
        - |
          objectName: $SECRET_NAME
          objectType: secret           # object types: secret, key or cert
          objectVersion: ""            # [OPTIONAL] object versions, default to latest if empty
    resourceGroup: $RESOURCE_GROUP     # [REQUIRED for version < 0.0.4] the resource group of the KeyVault
    subscriptionId: $SUBSCRIPTION_ID   # [REQUIRED for version < 0.0.4] the subscription ID of the KeyVault
    tenantId: $TENANT_ID               # the tenant ID of the KeyVault  
```
pod definition
```
kind: Pod
apiVersion: v1
metadata:
  name: nginx-secrets-store-env-var
  namespace: kube-infrastructure
  labels:
    aadpodidbinding: $IDENTITY_NAME
spec:
  containers:
    - name: nginx
      image: nginx
      volumeMounts:
      - name: secrets-store-inline
        mountPath: "/mnt/secrets-store"
        readOnly: true
      env:
      - name: SECRET_PASSWORD
        valueFrom:
          secretKeyRef:
            name: $SECRET_NAME
            key: password
  volumes:
    - name: secrets-store-inline
      csi:
        driver: secrets-store.csi.k8s.io
        readOnly: true
        volumeAttributes:
          secretProviderClass: "azure-kv-test-env-vars"
```

when trying to create the pod, the pod is stuck in `CreateContainerConfigError` state due to the first failure of secret volume mount and secret is not found.

Here's the log 
```
  Warning  FailedMount  2m42s                kubelet, aks-default-28105498-vmss000001  MountVolume.SetUp failed for volume "secrets-store-inline" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
  Normal   Pulling      76s (x8 over 2m40s)  kubelet, aks-vmss000001  Pulling image "nginx"
  Normal   Pulled       75s (x8 over 2m39s)  kubelet, aks-vmss000001  Successfully pulled image "nginx"
  Warning  Failed       75s (x8 over 2m39s)  kubelet, aks-vmss000001  Error: secret "test-aks-kv-int" not found
``` 
Need to find a way to get the pod ignoring `secret not found` issue till the mount is successfully done, not sure how !

The pod should be up and running with the secret exposed as ENV variable.

Azure Key Vault

**Environment:**

- Secrets Store CSI Driver version: 0.0.7
- Kubernetes version: 1.16.9

aramase:
- Uses `exec.CommandContext` when calling the provider binary. This ensures if the provider doesn't respond before the context timeout, then command times out and the process is killed. The `NodePublishVolume` will then return an error and also unmount the previously mounted `targetPath`. When `kubelet` retries the mount again with the `targetPath` for the pod, the `NodePublishVolume` will be retried as expected. 

Fixes #236 

### DIFF
--- a/pkg/secrets-store/nodeserver.go
+++ b/pkg/secrets-store/nodeserver.go
@@ -60,10 +60,19 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	var parameters map[string]string
 	var providerName string
 	var podName, podNamespace, podUID string
+	var targetPath string
+	var mounted bool
 	errorReason := FailedToMount
 
 	defer func() {
 		if err != nil {
+			// if there is an error at any stage during node publish volume and if the path
+			// has already been mounted, unmount the target path so the next time kubelet calls
+			// again for mount, entire node publish volume is retried
+			if targetPath != "" && mounted {
+				log.Infof("unmounting target path %s as node publish volume failed", targetPath)
+				ns.mounter.Unmount(targetPath)
+			}
 			ns.reporter.reportNodePublishErrorCtMetric(providerName, errorReason)
 			return
 		}
@@ -84,18 +93,18 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.InvalidArgument, "Volume attributes missing in request")
 	}
 
-	targetPath := req.GetTargetPath()
+	targetPath = req.GetTargetPath()
 	volumeID := req.GetVolumeId()
 	attrib := req.GetVolumeContext()
 	mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
 	secrets := req.GetSecrets()
 
-	mnt, err := ns.ensureMountPoint(targetPath)
+	mounted, err = ns.ensureMountPoint(targetPath)
 	if err != nil {
 		errorReason = FailedToEnsureMountPoint
 		return nil, status.Errorf(codes.Internal, "Could not mount target %q: %v", targetPath, err)
 	}
-	if mnt {
+	if mounted {
 		log.Infof("NodePublishVolume: %s is already mounted", targetPath)
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/secrets-store/nodeserver.go
```diff
@@ -60,10 +60,17 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	var targetPath string
 	errorReason := FailedToMount
 
 	defer func() {
 		if err != nil {
+			// if there is an error at any stage during node publish volume, we need
+			// to unmount the target path so that the next time kubelet calls again for
+			// mount, we retry the whole node publish volume
+			if targetPath != "" {
+				ns.mounter.Unmount(targetPath)
```
ritazh, line 72: should we first check if the path is mounted?

ritazh: We should add a test for this.

pkg/secrets-store/nodeserver_test.go
```diff
@@ -0,0 +1,105 @@
+/*
```
aramase, line 1: setting up initial unit test framework so we can expand the coverage in subsequent PRs

pkg/secrets-store/nodeserver.go
aramase: Added `mounted` boolean we can leverage before deciding if we need to unmount. The e2e test scenario would require the provider to respond only after the context deadline for this particular case.

ritazh: How about we add an e2e test for this to a long running periodic job?

PR #258, opened 2020-07-24
doc: add release doc and targets

### LINKED ISSUE #201: Add scripts to generate helm package and manifests
-  Add scripts to package and update index for helm during release

aramase:
- Adds docs for release process, some helper targets to run as part of generating the helm charts and release manifests.
- Adds `manifest_staging` dir to stage deployment manifest changes required for next release. There is no charts dir here, as new flags for helm charts can be added by wrapping them in conditionals similar to [this](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L54-L56). The helm charts are always used to test PRs and master.

Fixes #201 

/kind documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,14 +133,14 @@ e2e-teardown:
 .PHONY: install-driver
 install-driver:
 ifdef TEST_WINDOWS
-		helm install csi-secrets-store charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
+		helm install csi-secrets-store manifest_staging/charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
 			--set windows.image.pullPolicy="IfNotPresent" \
 			--set windows.image.repository=$(REGISTRY)/$(IMAGE_NAME) \
 			--set windows.image.tag=$(IMAGE_VERSION) \
 			--set windows.enabled=true \
 			--set linux.enabled=false
 else
-		helm install csi-secrets-store charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
+		helm install csi-secrets-store manifest_staging/charts/secrets-store-csi-driver --namespace default --wait --timeout=15m -v=5 --debug \
 			--set linux.image.pullPolicy="IfNotPresent" \
 			--set linux.image.repository="e2e/secrets-store-csi" \
 			--set linux.image.tag=$(IMAGE_VERSION) \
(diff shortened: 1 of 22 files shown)


docs/RELEASE.md
```diff
@@ -0,0 +1,73 @@
...
+    git checkout -b bump-version-<NEW VERSION>
+    git commit -m "Bump versions for <NEW VERSION"
+    git push <YOUR FORK>
+    ```
+   
+ 1. Once the PR is merged to master, the [prow job](https://testgrid.k8s.io/sig-auth-secrets-store-csi-driver#secrets-store-csi-driver-push-image) is triggered to build and push the new version to staging repo (`gcr.io/k8s-staging-csi-secrets-store/driver`)
+ 1. Once the prow job completes, follow the [instructions](https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#image-promoter) to promote the image to production repo
+    - Run generate script to append the new image to promoter manifest
+    ```bash
+    k8s.gcr.io/images/k8s-staging-csi-secrets-store/generate.sh > k8s.gcr.io/images/k8s-staging-csi-secrets-store/images.yaml
```
aramase, line 35: PR for adding generate script - https://github.com/kubernetes/k8s.io/pull/1053

Makefile
```diff
@@ -158,10 +158,10 @@ e2e-vault: install-driver
 # Update CRDs in Helm Chart
 release: manifests
 	cp config/crd/bases/* charts/secrets-store-csi-driver/templates
-	cp config/crd/bases/* deploy/
+	cp config/crd/bases/* manifest_staging/deploy/
```
ritazh, line 161: where do we use `make release`?

can these go under `manifests`?

do we also need `manifest_staging` for chart updates?

aramase: > where do we use make release?

`make release` is used in `promote-staging-manifest`

> do we also need manifest_staging for chart updates?

For charts, we can always add `semver` compare checks while adding new flags. The latest master chart is always used to run e2e tests. If we do maintain a copy of those charts in the `manifest_staging` dir, then we would need to run tests against those. For deploy manifests we need `manifest_staging` because the manifest are bound to the version, but with helm the semver compare does the trick. WDYT?

ritazh: The rationale behind adding the latest changes to `manifest_staging` instead of master is to ensure all configuration changes are released at the same time as the image in case there's any compatibility issues. e.g. a new flag. If a user grabs the chart from master, it should just work. I see your point about `semver` compare, but would be nice to not having to do that every time we make a chart update. As for e2es, we should test against `manifest_staging` chart.

aramase: Thank you! Makes sense. I'll add that.

Makefile
```diff
@@ -155,17 +155,14 @@ e2e-azure: install-driver
...
-	$(KUSTOMIZE) build config/rbac -o deploy/rbac-secretproviderclass.yaml
-
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=secretproviderclasses-role paths="./..." output:crd:artifacts:config=config/crd/bases
+	cp config/crd/bases/* manifest_staging/charts/secrets-store-csi-driver/templates
+	cp config/crd/bases/* manifest_staging/deploy/
+
+	# generate rbac-secretproviderclass
+	$(KUSTOMIZE) build config/rbac -o manifest_staging/deploy/rbac-secretproviderclass.yaml
```
ritazh, line 165: how does this change get into `manifest_staging/chart`

docs/RELEASE.md
```diff
@@ -0,0 +1,77 @@
...
+    git diff
+    ```
+    - Commit the changes and push to remote repository to create a pull request
+  1. Once the image promoter PR is merged, the image will be promoted to prod repo (`us.gcr.io/k8s-artifacts-prod/csi-secrets-store/driver`)
+  
+## Building and releasing
+
+1. Execute the release-patch target to generate patch. Give the semantic version of the release:
+    ```
+   make release-manifests NEWVERSION=v0.0.12
```
ritazh, line 48: If `make release-manifest` is only used when we cut a release, then you can have `promote-staging-manifest` call it so we only need to run a single command.

Makefile
aramase: Updated!

aramase:
@ritazh Updated the PR, PTAL!

PR #259, opened 2020-07-27
Add stripping sensitive information while logging the grpc request

### LINKED ISSUE #144: Add csi_secret
to filter sensitive information from logs
https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#handling-sensitive-information

helayoty:
Add stripping sensitive information while logging the grpc request.

Fixes #144 

### DIFF
--- a/pkg/csi-common/utils.go
+++ b/pkg/csi-common/utils.go
@@ -23,6 +23,8 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	pbSanitizer "github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
+
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 )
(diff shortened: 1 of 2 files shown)


pkg/csi-common/utils.go
```diff
@@ -95,7 +97,7 @@ func RunControllerandNodePublishServer(endpoint string, d *CSIDriver, cs csi.Con
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	log.Debugf("GRPC call: %s", info.FullMethod)
-	logRedactedRequest(req)
+	log.Debugf("GRPC request: %+v", pbSanitizer.StripSecretsCSI03(req).String())
```
aramase, line 100: `StripSecretsCSI03 ` doesn't work for CSI spec v1.0+  - https://github.com/kubernetes-csi/csi-lib-utils/blob/master/protosanitizer/protosanitizer.go#L49-L54. This driver uses `v1.1.0`, so we should switch to using `StripSecrets`

aramase: As `pbSanitizer.StripSecrets(req).String()` is a string, we should change the format specifier to `%s` instead of `%+v`

pkg/csi-common/utils.go
```diff
@@ -104,21 +106,3 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 	}
```
aramase, line 106: Can we also use the `protosanitizer` for the logging the GRPC response?

pkg/csi-common/utils.go
```diff
@@ -95,30 +97,12 @@ func RunControllerandNodePublishServer(endpoint string, d *CSIDriver, cs csi.Con
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	log.Debugf("GRPC call: %s", info.FullMethod)
-	logRedactedRequest(req)
+	log.Debugf("GRPC request: %+s", pbSanitizer.StripSecrets(req).String())
```
aramase, line 100: nit: Just `%s` for both the logs

PR #264, opened 2020-07-29
fix: use namespace for spc lookup + unit tests

### LINKED ISSUE #263: namespace lookup for SecretProviderClass not enforced in NodePublishVolume
`podNamespace` is set after fetching `SecretProviderClass` which means the namespace match isn't strictly enforced

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/d59cc202c99980546a68d354206cbb792f80ef19/pkg/secrets-store/nodeserver.go#L151

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/d59cc202c99980546a68d354206cbb792f80ef19/pkg/secrets-store/nodeserver.go#L133

**Environment:**

aramase:
- Sets the `podNamespace` env var for strict namespace enforcement
- Refactors `NodePublishVolume` to add more unit test for the `NodePublish` code path
- Updates logs to include pod name and pod namespace for debugging
- Adds unit tests 

Fixes #263 


pkg/secrets-store/nodeserver_test.go
```diff
@@ -84,14 +96,154 @@ func TestNodePublishVolume(t *testing.T) {
...
+					Spec: v1alpha1.SecretProviderClassSpec{
+						Provider:   "provider1",
+						Parameters: map[string]string{"parameter1": "value1"},
+					},
+				},
+			},
+			expectedErr: true,
+		},
+		{
+			name: "failed to invoke provider, unmounted to force retry",
```
ritazh, line 179: where do we test retry?

cmd/secrets-store-csi-driver/main.go
```diff
@@ -96,5 +98,13 @@ func main() {
 
 func handle() {
 	driver := secretsstore.GetDriver()
-	driver.Run(*driverName, *nodeID, *endpoint, *providerVolumePath, *minProviderVersion)
+	cfg, err := config.GetConfig()
```
ritazh, line 101: I recall we had to call `getClient()` to get the latest client to prevent from reading from cache. Is this no longer a concern?

test/bats/azure.bats
```diff
@@ -125,13 +126,70 @@ setup() {
   [[ "${result//$'\r'}" -eq "${SECRET_VALUE}" ]]
 
   result=$(kubectl get secret foosecret -o json | jq '.metadata.ownerReferences | length')
-  [[ "$result" -eq "2" ]]
+  [[ "$result" == "2" ]]
```
ritazh, line 129: i think `==` fails for some systems. can we make sure we consistently use `-eq` in case people need to run this locally?

test/bats/azure.bats
```diff
@@ -125,13 +126,70 @@ setup() {
...
+
+  run kubectl create secret generic secrets-store-creds --from-literal clientid=${AZURE_CLIENT_ID} --from-literal clientsecret=${AZURE_CLIENT_SECRET} -n test-ns
+  assert_success
+
+  envsubst < $BATS_TESTS_DIR/azure_v1alpha1_secretproviderclass_ns.yaml | kubectl apply -f -
+
+  cmd="kubectl wait --for condition=established --timeout=60s crd/secretproviderclasses.secrets-store.csi.x-k8s.io"
+  wait_for_process $WAIT_TIME $SLEEP_TIME "$cmd"
+
+  cmd="kubectl get secretproviderclasses.secrets-store.csi.x-k8s.io/azure-sync -o yaml | grep azure"
```
ritazh, line 156: why are we checking this in default ns? seems it's deleted on line 136

test/bats/azure.bats
```diff
@@ -125,13 +126,70 @@ setup() {
...
+  wait_for_process $WAIT_TIME $SLEEP_TIME "$cmd"
+
+  cmd="kubectl get secretproviderclasses.secrets-store.csi.x-k8s.io/azure-sync -n test-ns -o yaml | grep azure"
```
ritazh, line 159: is there also a negative test for referring to a spc in a different ns?

test/bats/azure.bats
aramase: `azure_v1alpha1_secretproviderclass_ns.yaml` creates 2 spc with the same name, one in `default` namespace and other in `test-ns`. This test ensures the spc in the same namespace as pod is matched and not the one in default namespace. The default namespace spc has an invalid provider - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/264/files#diff-b09039a0f14308e43877bb2112de35f1R6 which is a good sign to ensure the correct SPC is always matched

ritazh: ah ok. i missed that.

cmd/secrets-store-csi-driver/main.go
aramase: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L42-L52

```
// New returns a new Client using the provided config and Options.
// The returned client reads *and* writes directly from the server
// (it doesn't use object caches).  It understands how to work with
// normal types (both custom resources and aggregated/built-in resources),
// as well as unstructured types.
```

aramase:
@ritazh Updated the PR, PTAL!

pkg/secrets-store/nodeserver_test.go
aramase: Added another loop just to test retry for all the failed calls

test/bats/azure.bats
aramase: Added a negative test case that deploy pods in `negative-test-ns`. The mount will fail with 

```
  Warning  FailedMount  3s (x8 over 67s)  kubelet, aks-nodepool1-36438484-vmss000001  MountVolume.SetUp failed for volume "secrets-store-inline" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Unknown desc = failed to get secretproviderclass negative-test-ns/azure-sync, error: secretproviderclasses.secrets-store.csi.x-k8s.io "azure-sync" not found
```

cmd/secrets-store-csi-driver/main.go
ritazh: I was thinking about this one https://github.com/kubernetes-sigs/controller-runtime/issues/403
But I dont think we have multiple controllers updating the same objects now.

aramase: Right, we don't have any update operations on `spc`. There is only `GET` for `spc` and `CREATE` for `spcpodstatus`

test/bats/vault.bats
```diff
@@ -188,7 +189,79 @@ EOF
...
+  envsubst < $BATS_TESTS_DIR/nginx-deployment-synck8s.yaml | kubectl apply -n test-ns -f -
+
+  cmd="kubectl wait --for=condition=Ready --timeout=60s pod -l app=nginx -n test-ns"
+  wait_for_process $WAIT_TIME $SLEEP_TIME "$cmd"
+}
+
+@test "Test Namespaced scope SecretProviderClass - Sync with K8s secrets - read secret from pod, read K8s secret, read env var, check secret ownerReferences" {
+  POD=$(kubectl get pod -l app=nginx -n test-ns -o jsonpath="{.items[0].metadata.name}")
+  result=$(kubectl exec -n test-ns -it $POD -- cat /mnt/secrets-store/foo)
+  [[ "$result" == "hello" ]]
```
ritazh, line 227: can you pls update all these `==` to `-eq`

aramase: I think we should use `==` for string comparisons. Using `-eq` returns 0 even when the strings don't match. 

```
#!/usr/bin/env bats

export result="hel"

@test "testing with eq" {
    [[ "$result" -eq "hello" ]]
}

@test "testing with other" {
    [[ "$result" == *"hello"* ]]
}

➜ bats test.bats -t
1..2
ok 1 testing with eq
not ok 2 testing with other
# (in test file /Users/anishramasekar/Desktop/test.bats, line 9)
#   `@test "testing with other" {' failed
```

test/bats/vault.bats
```diff
@@ -188,7 +189,79 @@ EOF
 @test "Sync with K8s secrets - delete deployment, check secret is deleted" {
   run kubectl delete -f $BATS_TESTS_DIR/nginx-deployment-synck8s.yaml
   assert_success
+
+  run kubectl delete -f $BATS_TESTS_DIR/vault_synck8s_v1alpha1_secretproviderclass.yaml
+  assert_success
+
   sleep 20
   result=$(kubectl get secret | grep foosecret | wc -l)
   [[ "$result" == "0" ]]
```
ritazh, line 198: As discussed, lets use `-eq` for numeric and `==` for string.

cmd/secrets-store-csi-driver/main.go
aramase: Adding the links here for future ref 

https://github.com/kubernetes-sigs/controller-runtime/blob/cea989b02ed02254d751ad30471c40bacd5888fc/pkg/client/client_cache.go#L88-L96
    
newResource always creates a new client - https://github.com/kubernetes-sigs/controller-runtime/blob/cea989b02ed02254d751ad30471c40bacd5888fc/pkg/client/client_cache.go#L51-L68

aramase:
/hold

(for windows tests to pass)

PR #276, opened 2020-08-05
test: move tests to subdir for provider

aramase:
- Moves test manifests into sub dir for each provider
- Removes `nginx-pod-secrets-store-inline-volume.yaml` which is no longer being used
- Adds windows job template for driver windows test

Fixes #


test/bats/tests/azure/job_templates/kubernetes_windows.json
```diff
@@ -0,0 +1,55 @@
...
+            "enableTelemetry": true
+        },
+        "orchestratorProfile": {
+            "orchestratorType": "Kubernetes",
+            "orchestratorRelease": "1.18",
+            "kubernetesConfig": {
+                "addons": [
+                    {
+                        "name": "csi-secrets-store",
+                        "enabled": true
```
ritazh, line 14: don't you want this to be `false`?

aramase: Yeah, updated!

PR #277, opened 2020-08-05
[WIP] feat: compute SHA for data and patch if data changed

aramase:
- Generates `SHA` for existing secret data and expected secret data. If the SHA doesn't match, then replace the contents with the expected content. This ensures when pods are deleted by `RollingUpdate` but secret isn't deleted, then the contents are kept up to date. 

Fixes #

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -130,51 +130,64 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(req ctrl.Request) (ct
 			errs = append(errs, fmt.Errorf("data is empty at index %d for secret %s", idx, secretObj.SecretName))
 			continue
 		}
-		exists, err := r.secretExists(ctx, secretObj.SecretName, req.Namespace)
+		exists, currentData, err := r.secretExists(ctx, secretObj.SecretName, req.Namespace)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


connor4312:
Any updates on this? Currently whenever I add a secret I need to manually delete the secret and the csi pods (sometimes more than once) before the change sticks. This is a very rough experience, I would love to see it fixed 🙂

aramase:
> Any updates on this? Currently whenever I add a secret I need to manually delete the secret and the csi pods (sometimes more than once) before the change sticks. This is a very rough experience, I would love to see it fixed 🙂

This will be included in the auto rotation feature that's being added for the next release v0.0.14.

connor4312:
Sounds good. Since it's been a couple months now, do you have steps that I can use to force secrets to be refreshed in the meantime? It seems that even deleting the pods and secret now (or sometimes) doesn't cause the secret to be refreshed. This is super frustrating and makes it hard to use keyvault in AKS 🙁

connor4312:
Ah, actually, in the case where deleting didn't work, it seems like the access policy on my keyvault had gone missing somehow. So that's my fault, but reliable steps to refresh secrets would still be appreciated!

aramase:
@connor4312 Sure, I'll add a doc with the current limitation and will update that as in when we release the auto-rotation feature.

aramase:
/close 

As changes will be part of #303

PR #303, opened 2020-08-31
feat: add rotation reconciler

### LINKED ISSUE #125: Automatic secrets reloading upon update in the external secrets store
There are many use-cases when secret/certificate is being updated in the KeyVault. The most prominent use-case is the rotation flow. Fortunately, KeyVault takes care of certificates and Azure Storage SAS rotations. When new secret/certificate has been created, the running pod will not start to utilize it till the restart. Only at restart the tmpfs will be created with renewed secrets. 
In most cases (but not all cases) the old secret will continue to work and there is a plenty of time for pod being restarted.
Everything goes complicated when there are many services that utilize many secrets. It's pretty normal situation when single cluster serves hundreds of services and thousands of secrets. 
Synchronization at scale brings a lot of mess into cluster operation.

I would like to request a new feature to secrets store to reload secrets upon update. Since many client's implementations do not support live reload, the pod restart will be the ideal solution.

aramase:
- Adds a new rotation reconciler that rotates secret contents mounted in the pod and synced as K8s secrets. The reconciler is invoked periodically based on an interval which can be configured using `--rotation-poll-interval`.
- The rotation reconciler can be enabled by setting `--enable-secret-rotation=true`. This is an `[alpha]` that's not enabled by default.
- Requires the provider to support GRPC server
- Adds support for a patcher that periodically reconciles `ownerReferences` in the secret. 
- Adds scripts for generating clientset, informers and listers using `code-generator`
- Adds generated clientset, informers and listers for `SecretProviderClass` and `SecretProviderClassPodStatus`

This PR is built on top of #280 

`Revised design` section in the design doc - [Secrets Store CSI Driver - Rotation](https://docs.google.com/document/d/1RGT0vmeUnN71n_u5fZKsSCa2YQpGw99rfGN9RlFMgHs/edit?usp=sharing)

Fixes #125 

**Attribution:**
1. The `hack/verify-codegen.sh` and `verify/update-codegen.sh` have been taken from https://github.com/kubernetes-sigs/service-apis project


controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -19,49 +19,136 @@ package controllers
...
 	client.Client
+	Mutex  *sync.Mutex
 	Log    *log.Logger
 	Scheme *runtime.Scheme
 	NodeID string
 	Reader client.Reader
 	Writer client.Writer
 }
 
+func (r *SecretProviderClassPodStatusReconciler) RunPatcher(stopCh <-chan struct{}) {
```
aramase, line 57: The `Patcher` doesn't need to part of this PR. I've included it as an optimization based on load tests. If required, I can stub this change out from this PR and create a new PR to just add the `Patcher` to keep reviews easy.

pkg/k8s/pod.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// GetWithKey returns pod with key from the informer cache
+func (pl *PodLister) GetWithKey(key string) (*v1.Pod, error) {
+	p, exists, err := pl.GetByKey(key)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, fmt.Errorf("pod not found in informer cache")
```
ritazh, line 38: Should we create some errorReasons for these so we can get metrics on them?

pkg/rotation/reconciler.go
```diff
@@ -0,0 +1,346 @@
...
+		case <-ticker.C:
+			log.Infof("rotation reconcile started")
+			// The spc pod status informer is configured to do a filtered list watch of spc pod statuses
+			// labeled for the same node as the driver. LIST will only return the filtered results.
+			spcpsList, err := r.store.ListSecretProviderClassPodStatus()
+			if err != nil {
+				log.Errorf("[rotation] failed to list secret provider class pod status for node, err: %+v", err)
+				continue
+			}
+			for _, spcps := range spcpsList {
```
ritazh, line 120: What happens if this loop takes longer than the rotation poll interval?

pkg/rotation/reconciler.go
```diff
@@ -0,0 +1,346 @@
...
+	oldObjectVersions := make(map[string]string)
+	for _, obj := range spcps.Status.Objects {
+		oldObjectVersions[obj.ID] = obj.Version
+	}
+
+	providerClient, err := r.getProviderClient(string(spc.Spec.Provider))
+	if err != nil {
+		return fmt.Errorf("failed to create provider client, err: %+v", err)
+	}
+	newObjectVersions, _, err := providerClient.MountContent(ctx, string(paramsJSON), string(secretsJSON), spcps.Status.TargetPath, string(permissionJSON), oldObjectVersions)
```
ritazh, line 204: After the new object version has been mounted, if something fails before the spcps is updated, then what is the mechanism that triggers the reconcile again? the interval? Right now I see any error returned from this function gets logged, can we requeue as well so another reconcile loop gets triggered?

controllers/secretproviderclasspodstatus_controller.go
ritazh: Are we patching here periodically in case the ownerReference gets removed from the synced secrets?

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -19,49 +19,136 @@ package controllers
...
+	log.Debugf("patcher started")
+	r.Mutex.Lock()
+	defer r.Mutex.Unlock()
+
+	ctx := context.Background()
+	spcPodStatusList := &v1alpha1.SecretProviderClassPodStatusList{}
+	spcMap := make(map[string]v1alpha1.SecretProviderClass)
+	secretOwnerMap := make(map[types.NamespacedName][]*v1alpha1.SecretProviderClassPodStatus)
+	// get a list of all spc pod status that belong to the node
+	err := r.Reader.List(ctx, spcPodStatusList, r.ListOptionsLabelSelector())
```
ritazh, line 83: Instead of calling list all the time, can we create a spcMap cache and update the cache on spcps reconcile? Then the Patcher can rely on the spcMap cache instead.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -19,49 +19,136 @@ package controllers
...
+func (r *SecretProviderClassPodStatusReconciler) ListOptionsLabelSelector() client.ListOption {
+	return client.MatchingLabels(map[string]string{
+		v1alpha1.InternalNodeLabel: r.NodeID,
+	})
+}
+
 // +kubebuilder:rbac:groups=secrets-store.csi.x-k8s.io,resources=secretproviderclasspodstatuses,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=secrets-store.csi.x-k8s.io,resources=secretproviderclasspodstatuses/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=secrets-store.csi.x-k8s.io,resources=secretproviderclasses,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
```
ritazh, line 146: I think this is only needed when rotation is enabled right? ie store.GetPod is only used by the rotation reconciler.

pkg/rotation/reconciler.go
```diff
@@ -0,0 +1,346 @@
...
+	pod, err := r.store.GetPod(podName, podNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to get pod %s/%s, err: %+v", podNamespace, podName, err)
+	}
+
+	// check if the volume pertaining to the current spc is using nodePublishSecretRef for
+	// accessing external secrets store
+	var nodePublishSecretRef *v1.LocalObjectReference
+	for _, vol := range pod.Spec.Volumes {
+		if vol.CSI.Driver != "secrets-store.csi.k8s.io" {
```
ritazh, line 158: do we need to check if the volume is csi here?

controllers/secretproviderclasspodstatus_controller.go
aramase: Yeah, this patcher adds the `ownerReferences` to the secret in batches. Periodically checks to see if any of the required `ownerRefs` are missing and adds them.

controllers/secretproviderclasspodstatus_controller.go
aramase: There is already an `spcMap` cache maintained in the patcher. The `List` call here shouldn't be expensive as it's List from the cache. Is the suggestion to maintain a cache map for the SPCPS too?

pkg/rotation/reconciler.go
aramase: `vol.CSI` checks only if the CSI volume is present and is for the `secrets-store.csi.k8s.io` driver.

pkg/rotation/reconciler.go
aramase: Yeah, right now it's controlled by the interval.

controllers/secretproviderclasspodstatus_controller.go
aramase: Yeah, this permission is required only for rotation reconciler. We added a separate role for secrets sync feature, but I don't think we want to do the same for pods too?

pkg/k8s/pod.go
aramase: I've added the errorReason to the top level reconcile method. I think the clients should still be generic and reused across other packages.

pkg/rotation/reconciler.go
aramase: I've updated this part to add the objects to the queue which are then processed by the worker. But even without the change, the `<-ticker.C` will be blocked until it's read from the channel. So if the loop takes longer than the rotation poll interval, the last tick will still be in the channel and until it's read from the channel, subsequent ticks will be dropped.

controllers/secretproviderclasspodstatus_controller.go
aramase: Added a worker queue as part of this commit - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/303/commits/a6227089818fa7714cd584b36620986947469f2c. Let me know what you think.

controllers/secretproviderclasspodstatus_controller.go
ritazh: Definitely interested in other's opinions on this. IMO if we are following the principle of least privilege, then this permission on pods is only needed if user wants to support rotation. otherwise, it's not needed.

pkg/rotation/reconciler.go
ritazh: Thanks for adding the queue! maybe as a followup, let's add duration for each run to our metrics. Would be useful for debugging and performance tests.

pkg/rotation/reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		errorReason = internalerrors.PodNotFound
+		return fmt.Errorf("failed to get pod %s/%s, err: %+v", podNamespace, podName, err)
+	}
+
+	// check if the volume pertaining to the current spc is using nodePublishSecretRef for
+	// accessing external secrets store
+	var nodePublishSecretRef *v1.LocalObjectReference
+	for _, vol := range pod.Spec.Volumes {
+		if vol.CSI == nil {
```
ritazh, line 190: +1

pkg/rotation/reconciler.go
ritazh: on error, can we requeue as well so another reconcile loop gets triggered?

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -259,11 +300,28 @@ func (r *SecretProviderClassPodStatusReconciler) patchSecretWithOwnerRef(ctx con
...
-		return err
+	needsPatch := false
+
+	secretOwnerMap := make(map[string]types.UID)
+	for _, or := range secret.GetOwnerReferences() {
+		secretOwnerMap[or.Name] = or.UID
+	}
+
+	for i := range spcPodStatus {
+		if _, exists := secretOwnerMap[spcPodStatus[i].Name]; exists {
```
ritazh, line 311: This ensures all current spcps are added to the secret's ownerReferences if missing. What about when an old spcps ownerReference needs to be removed?

controllers/secretproviderclasspodstatus_controller.go
ritazh: > Added a worker queue as part of this commit 

Isn't the worker queue for rotation reconciler? Patcher is for reconciling ownerReference of SecretProviderClassPodStatus right?

> The List call here shouldn't be expensive as it's List from the cache

Do we want to get the latest spcPodStatusList and there spcs from the api server or is getting from the cache ok?

controllers/secretproviderclasspodstatus_controller.go
aramase: The spcps contains the pod as owner reference. When the pod is deleted, the corresponding spcps is deleted as part of cascading delete and once spcps is deleted, garbage collector will remove the deleted spcps from the owner reference list

controllers/secretproviderclasspodstatus_controller.go
aramase: Ah, the comments got mixed up. `Added a worker queue as part of this commit - a622708. Let me know what you think.` was in response to the previous comment. 

> Isn't the worker queue for rotation reconciler? Patcher is for reconciling ownerReference of SecretProviderClassPodStatus right?

Yes, that's correct. 

> Do we want to get the latest spcPodStatusList and there spcs from the api server or is getting from the cache ok?

I would consider getting from the cache here a safe operation. The cache informers get events on the watch channels as there are changes in the resources and repopulate the cache frequently. The patcher is configured with an aggressive interval of `5s` which gives enough time for the caches to be populated with the current set of resources.

pkg/rotation/reconciler.go
aramase: [processNextItem](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/303/files#diff-1c6cad8993ddd7d5baf508039d01ff3bR382-R401) picks up the object from the worker queue and reconciles it. In case of errors we [handleError](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/303/files#diff-1c6cad8993ddd7d5baf508039d01ff3bR403-R410) which requeues the spcps for reconcile after 10s.

controllers/secretproviderclasspodstatus_controller.go
ritazh: That makes sense. Can we add a test for pod deletion to ensure the ownerReference is always cleaned up?

aramase: This test helps us validate the cascade delete today - https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/test/bats/azure.bats#L172-L199

ritazh: I don't see a test that deletes the pod from a list of ownerReferences of a secret used by multiple owners.

aramase: PR for testing secret with multiple owners - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/309. The single owner is a subset of this new test.

tam7t (approved):

aramase:
> Is this needed because each node needs to query to find out what pods are still active with secrets? Can that not be determined by the secretproviderclasspodstatuses resource? I think my biggest confusion here was understanding why the permission was needed.

@tam7t So, this permission is required to find out if the pod is using a `NodePublishSecretRef` for volume attributes. [NodePublishSecretRef](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html#node-publish-secret) can be configured by the user for setting up credentials to access the external secrets store. The content of the secret is the data for the secrets in the `MountRequest` payload. The `SecretProviderClassPodStatus` is all we check for all the required info. The pod GET permission is to only check if the pod is using `NodePublishSecretRef` for the volume and then get the secret data.

cmd/secrets-store-csi-driver/main.go
```diff
@@ -46,6 +49,8 @@ var (
 	// grpcSupportedProviders is a ; separated string that can contain a list of providers. The reason it's a string is to allow scenarios
 	// where the driver is being used with 2 providers, one which supports grpc and other using binary for provider.
 	grpcSupportedProviders = flag.String("grpc-supported-providers", "", "set list of providers that support grpc for driver-provider [alpha]")
+	enableSecretRotation   = flag.Bool("enable-secret-rotation", false, "Enable secret rotation feature [alpha]")
+	rotationPollInterval   = flag.Duration("rotation-poll-interval", 60, "Secret rotation poll interval duration")
```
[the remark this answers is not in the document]
jasonodonnell: 💯 on the SPC. The lifecycle of a secret is configurable in Vault, so secrets would have different rotation periods.  This would be a nice to have but not absolutely required since we could add to the Vault Agent sidecar to the secret store pod with caching enabled. The agent would renew secrets, know if secrets are still valid or serve new secrets if they're expired. 

We could set a very aggressive poll interval and lean on the sidecar to make it work with Vault.

controllers/secretproviderclasspodstatus_controller.go
ritazh: As discussed in last week's community call, "get pod" is acceptable as a base-level rbac for this driver as it does not allow escalation between node workloads.

cmd/secrets-store-csi-driver/main.go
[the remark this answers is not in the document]
aramase: > It might be nice for this to be a configurable option in the SPC CRD instead of a flag on the controller, but this seems good to start.

@tam7t What is the use case for this? Is the idea that secrets have different lifetime period? Having a global poll interval flag makes it consistent across all the SPCs. The provider would get the current version and the latest version. Only if latest != current would a write to the fs be required.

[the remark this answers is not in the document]
aramase: I've opened an issue for this - https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/322

skri547:
@aramase  Can you let us know how to use this feature. I am currently facing the same issue, I had to restart the pod to get the new value. If you let me know how to implement this that would be great.
Thanks

aramase:
@skri547 The docs for this project are available here: https://secrets-store-csi-driver.sigs.k8s.io

Rotation documentation: https://secrets-store-csi-driver.sigs.k8s.io/topics/secret-auto-rotation.html

skri547:
@aramase  Thanks for the quick reply, I am unable to know where to enable the --enable-secret-rotation feature gate for the secrets-store container in the Secrets Store CSI Driver pods ? Is that in deployment of nginx (example) or can you provide that detail?
Thanks

aramase:
@skri547 If you're using helm charts to install the provider, then you can set `enableSecretRotation=true` while deploying the driver. The same applies to helm charts from the provider packaging the driver as dependency. For example https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/charts/csi-secrets-store-provider-azure/values.yaml#L116-L117.

If you're installing the driver using the yaml files, then enable it by setting it to `true` in the args for secrets-store container here
https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/73b7af5ee4615ccf665810139ece3cd092a3e9e8/deploy/secrets-store-csi-driver.yaml#L51-L52

skri547:
Apart from this, is anything needed to make sure this is working?

skri547:
@aramase I deployed as you said and got the below error 

`Warning  FailedMount  95s (x5 over 10m)  kubelet            Unable to attach or mount volumes: unmounted volumes=[secrets-store-inline], unattached volumes=[secrets-store-inline default-token-478xx]: timed out waiting for the condition
  Warning  FailedMount  59s (x8 over 11m)  kubelet            MountVolume.SetUp failed for volume "secrets-store-inline" : rpc error: code = Unknown desc = failed to mount secrets store objects for pod imasservices/busybox-secrets-store-inline, err: rpc error: code = Unknown desc = failed to mount objects, error: failed to get keyvault client: failed to get key vault token: nmi response failed with status code: 404, err: <nil>`

skri547:
I only have one secret in key-vault and I have added that in SecretProviderClass aswell

aramase:
@skri547 If this is specific question for the Azure provider, please open a new issue [here](https://github.com/Azure/secrets-store-csi-driver-provider-azure/issues) with all the relevant information and logs.

PR #308, opened 2020-09-11
feat: add unit tests for secrets-store pkg

### LINKED ISSUE #132: Add more ut for NodePublishVolume and NodeUnPublishVolume
- [ ] increase test coverage for bad requests
- [ ] increase test coverage to check for diff returned error types
- [ ] increase test coverage to check for IsLikelyNotMountPoint

irbekrm:

Adds more tests for NodePublishVolume and NodeUnpublishVolume functions

Fixes #132 

I think it would be great if it was possible to also test success scenario for NodePublishVolume. At the moment the code that executes the provider binary with args runs directly in the NodePublishVolume function, but maybe if could be factored out?

### DIFF
--- a/pkg/secrets-store/mocks/stats_reporter_mock.go
+++ b/pkg/secrets-store/mocks/stats_reporter_mock.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mocks // import sigs.k8s.io/secrets-store-csi-driver/pkg/secrets-store/mocks
+
+type FakeReporter struct {
+	reportNodePublishCtMetricInvoked        int
+	reportNodeUnPublishCtMetricInvoked      int
+	reportNodePublishErrorCtMetricInvoked   int
+	reportNodeUnPublishErrorCtMetricInvoked int
+	reportSyncK8SecretCtMetricInvoked       int
+	reportSyncK8SecretDurationInvoked       int
+}
+
+func NewFakeReporter() *FakeReporter {
+	return &FakeReporter{}
+}
+
+func (f *FakeReporter) ReportNodePublishCtMetric(provider string) {
+	f.reportNodePublishCtMetricInvoked++
+}
+
+func (f *FakeReporter) ReportNodeUnPublishCtMetric() {
+	f.reportNodeUnPublishCtMetricInvoked++
+}
+
+func (f *FakeReporter) ReportNodePublishErrorCtMetric(provider, errType string) {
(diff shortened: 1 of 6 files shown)


aramase:
> I think it would be great if it was possible to also test success scenario for NodePublishVolume. At the moment the code that executes the provider binary with args runs directly in the NodePublishVolume function, but maybe if could be factored out?


That's a good idea. I was thinking the same and started doing that as part of this https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/280 which was merged recently. Take a look when you get a chance.

aramase:
/retitle feat: add unit tests for secrets-store pkg

aramase (commented):
@irbekrm Thank you for getting started on this. PR looks great 🎉 

Just one small nit. Can we create a `mocks` dir in `pkg/secrets-store` and move `fakes.go` to that dir? We can then consolidate all test mocks we add in the future. Could we also rename `fakes.go` to `stats_reporter_mock.go`?

irbekrm:
Thanks @aramase ! Will look into that tomorrow morning

irbekrm:
Hi @aramase thank you for the review. I have put the stats reporter mock in `pkg/secrets-store/mocks/stats_reporter_mock.go`. I exported the `StatsReporter` interface because the mock in now in a different package. I then also moved the stats reporter functionality into `pkg/secrets-store/internal/statsreporter` package to keep it internal to `pkg/secrets-store`.

aramase:
> Hi @aramase thank you for the review. I have put the stats reporter mock in pkg/secrets-store/mocks/stats_reporter_mock.go. I exported the StatsReporter interface because the mock in now in a different package. I then also moved the stats reporter functionality into pkg/secrets-store/internal/statsreporter package to keep it internal to pkg/secrets-store.

For `StatsReporter` I was planning to keep the implementation for the interface inside each package instead of a global `statsreporter` package. For instance, I'm adding a new `stats_reporter.go` for the rotation reconciler that I'm adding as part of auto-rotation feature - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/303/files#diff-d1f4e657fb85fe2e0870aac577505123R14-R67. This ensures isolation for each implementation. What are your thoughts on that? 

Also, could you rebase the PR and squash the commits?

irbekrm:
Hi @aramase, thank you for the comment! 

About the `StatsReporter`- after I moved the mocks to a separate package I needed to export the `StatsReporter`'s methods so that the interface could still be mocked. That would have made the `StatsReporter`'s functionality exported externally and available in all other packages, so I put the reporter into `/pkg/secrets-store/internal/statsreporter` so the methods were still internal to that package path and not available externally. But admittedly that is a bit clunky, each package that would want to add a stats reporter would need to put it into an `internal/statsreporter` subpackage for consistency. 

Alternatively, I guess maybe the methods could just be public? Or maybe it is not crucial to test them being called from `NodePublishVolume`/`NodeUnpublishVolume`?

Thank you for reminding to squash and rebase- I have done that now.

aramase:
> Alternatively, I guess maybe the methods could just be public? Or maybe it is not crucial to test them being called from NodePublishVolume/NodeUnpublishVolume?

I really like how we are testing it in this PR. Keeping the `stats_reporter.go` in the same package and making it public sounds good! We can go with that approach.

irbekrm:
Thanks @aramase! 
I agree and have made that change. (Since I'm not that familiar with this codebase I wasn't sure what would make most sense)

aramase (commented):
@irbekrm The PR looks great! 🎉 

lgtm

PR #323, opened 2020-09-28
feat: add event recorder for secrets creation and rotation

### LINKED ISSUE #307: generate events for secret creation and rotation
- Generate events for pod to signal secret creation and failures
- We can also use this event recorder to publish events when the rotation reconciler completes updating the content for the pod. 

**Environment:**

aramase:
- Adds event recorder to generate events for secret creation failure/ when mounted contents are rotated
- Adds SHA check to determine if updating secret is required
- 

Fixes #307 

```bash
  Normal  MountRotationComplete   8s    csi-secrets-store-rotation  successfully rotated mounted contents for spc default/azure
  Normal  SecretRotationComplete  8s    csi-secrets-store-rotation  successfully rotated K8s secret foosecret
```

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -19,13 +19,21 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
 
+	"k8s.io/client-go/kubernetes"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"k8s.io/client-go/tools/record"
+
 	log "github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/secrets-store-csi-driver/apis/v1alpha1"
+	"sigs.k8s.io/secrets-store-csi-driver/pkg/client/clientset/versioned/scheme"
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/util/fileutil"
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/util/secretutil"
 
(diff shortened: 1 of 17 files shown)


aramase:
/test pull-secrets-store-csi-driver-e2e-azure

(retrying now that AAD failures should be resolved)

aramase:
@ritazh @tam7t @tomhjp PTAL when you get a chance!

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -308,7 +350,11 @@ func (r *SecretProviderClassPodStatusReconciler) patchSecretWithOwnerRef(ctx con
 		Name:      name,
 	}
 	err := r.Client.Get(ctx, secretKey, secret)
-	if err != nil && !errors.IsNotFound(err) {
+	if apierrors.IsNotFound(err) {
```
ritazh, line 353: should we check for nil first?

pkg/rotation/reconciler.go
```diff
@@ -44,19 +48,24 @@ import (
...
+
 	"sigs.k8s.io/secrets-store-csi-driver/apis/v1alpha1"
 	secretsStoreClient "sigs.k8s.io/secrets-store-csi-driver/pkg/client/clientset/versioned"
-
-	log "github.com/sirupsen/logrus"
 )
 
 const (
-	permission os.FileMode = 0644
+	permission                      os.FileMode = 0644
```
ritazh, line 64: nit: space

pkg/rotation/reconciler.go
```diff
@@ -389,11 +415,11 @@ func (r *Reconciler) processNextItem() bool {
 	spcps, err := r.store.GetSecretProviderClassPodStatus(key.(string))
 	if err != nil {
 		log.Errorf("failed to get spc pod status, error: %+v", err)
-		r.handleError(err, key)
+		return true
```
ritazh, line 418: what's the reason for this change?

controllers/secretproviderclasspodstatus_controller.go
aramase: *logically* `apierrors.IsNotFound` returns false if error is nil or error is anything other than not found. So we can do the check even without the nil check first.

But I think the code might be more readable with the `err != nil` check. So I've updated the PR.

pkg/rotation/reconciler.go
aramase: Did some research and updated to use rate limited queueing to still allow retries when the `SecretProviderClassPodStatus` is not found. Also added a comment to describe why we are making the changes.

tam7t (approved):
LGTM

pkg/rotation/reconciler.go
```diff
@@ -286,50 +314,65 @@ func (r *Reconciler) reconcile(ctx context.Context, spcps *v1alpha1.SecretProvid
 		}, updateFn); err != nil {
 			return fmt.Errorf("failed to update spc pod status, err: %+v", err)
```
ritazh, line 315: Do we also want to generate an event for this failure too?

aramase: Yeah, added an event for failing to spcps with the latest versions.

PR #327, opened 2020-09-30
feat: add trimspace to sanitize yaml fields

aramase:
- Add `strings.TrimSpace` for all fields obtained as string from manifests
- Adds positive test cases for `GetSecretData`

Fixes #

### DIFF
--- a/pkg/util/secretutil/secret.go
+++ b/pkg/util/secretutil/secret.go
@@ -23,6 +23,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	"sigs.k8s.io/secrets-store-csi-driver/apis/v1alpha1"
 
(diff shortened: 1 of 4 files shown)


pkg/util/secretutil/secret.go
```diff
@@ -157,7 +158,7 @@ func GetSecretData(secretObjData []*v1alpha1.SecretObjectData, secretType corev1
 		if len(data.Key) == 0 {
 			return datamap, fmt.Errorf("key in secretObjects.data is empty")
 		}
-		file, ok := files[data.ObjectName]
+		file, ok := files[strings.TrimSpace(data.ObjectName)]
```
ritazh, line 161: Do we also need to trim space for `data.Key`?

aramase: Updated the PR!

pkg/util/secretutil/secret.go
```diff
@@ -151,27 +152,30 @@ func ValidateSecretObject(secretObj v1alpha1.SecretObject) error {
...
+
+		if len(objectName) == 0 {
 			return datamap, fmt.Errorf("object name in secretObjects.data")
 		}
-		if len(data.Key) == 0 {
+		if len(dataKey) == 0 {
 			return datamap, fmt.Errorf("key in secretObjects.data is empty")
 		}
-		file, ok := files[data.ObjectName]
+		file, ok := files[strings.TrimSpace(objectName)]
```
ritazh, line 164: why do you need to trim again here?

aramase: s/`data.ObjectName`/`objectName` 🤦 

Updated it!

PR #351, opened 2020-10-19
deploy: add gcp as grpcSupportedProviders by default

tam7t:

The gcp plugin only works with `--grpc-supported-providers`, so adding it to the template to set by default.

### DIFF
--- a/manifest_staging/charts/secrets-store-csi-driver/values.yaml
+++ b/manifest_staging/charts/secrets-store-csi-driver/values.yaml
@@ -120,7 +120,7 @@ syncSecret:
 minimumProviderVersions:
 
 ## ; delimited list of providers that support grpc for driver-provider [alpha]
-grpcSupportedProviders:
+grpcSupportedProviders: gcp;
 
 ## Enable secret rotation feature [alpha]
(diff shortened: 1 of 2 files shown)


manifest_staging/charts/secrets-store-csi-driver/values.yaml
```diff
@@ -120,7 +120,7 @@ syncSecret:
 minimumProviderVersions:
 
 ## ; delimited list of providers that support grpc for driver-provider [alpha]
-grpcSupportedProviders:
+grpcSupportedProviders: gcp;
```
ritazh, line 123: Should we include all the grpc supported providers here?

aramase: The reason we haven't included is because azure provider versions prior to `0.0.8` still work with the current driver. If the user wishes to use the latest provider they set the args.

ritazh: That makes sense!

aramase (commented):
/lgtm

PR #373, opened 2020-11-05
fix: handle pod termination during reconcile

aramase:
Checks if the pod is terminating while inside the reconcile or rotation loop. If the pod is terminating, then return.

Fixes #

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -209,6 +209,22 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Obtain the full pod metadata. An object reference is needed for sending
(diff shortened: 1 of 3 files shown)


controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -209,6 +209,22 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// Obtain the full pod metadata. An object reference is needed for sending
+	// events and the UID is helpful for validating the SPCPS TargetPath.
+	pod := &v1.Pod{}
```
aramase, line 214: Moving the get pod to the start of the loop in order to avoid additional API calls if the pod is being terminated

tam7t: SGTM

tam7t (approved):

PR #383, opened 2020-11-16
feat: create target path in node publish

### LINKED ISSUE #382: target path will not be created before NodePublish for 1.20+
As added in the changelog -
```
In a future release, kubelet will no longer create the CSI NodePublishVolume target directory, in accordance with the CSI specification. CSI drivers may need to be updated accordingly to properly create and process the target path. (#75535) [SIG Storage]
```

Reference announcement - https://groups.google.com/g/container-storage-interface-drivers-announce/c/GgtP9kiv5Qc/m/CO2yyOJvAAAJ

We need to check if the target path exists, if not create it in the CSI driver

**Environment:**

aramase:
Starting from release 1.20, kubelet will no longer create the target path. The CSI driver is required to create the target path. 
- Update to the latest version of `csi-test` that checks the target path is created by the CSI driver
- Update to the latest version of `csi-spec`

Fixes #382 


test/sanity/sanity_test.go
```diff
@@ -36,10 +37,32 @@ func TestSanity(t *testing.T) {
...
-		StagingPath: stagePath,
+		TargetPath:  os.TempDir() + "/csi-target",
+		StagingPath: os.TempDir() + "/csi-staging",
 		Address:     endpoint,
+		CreateTargetDir: func(targetPath string) (string, error) {
+			targetPath = path.Join(tmpPath, targetPath)
+			return targetPath, createTargetDir(targetPath)
+		},
+		RemoveTargetPath: func(targetPath string) error {
+			return os.RemoveAll(targetPath)
```
tam7t, line 50: Should this be `os.RemoveAll(tmpPath)`?

test/sanity/sanity_test.go
```diff
@@ -15,6 +15,9 @@ package sanity
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path"
```
tam7t, line 20: you may want to use `filepath` instead of `path` package

go.mod
```diff
@@ -4,12 +4,12 @@ go 1.15
...
 	github.com/blang/semver v3.5.0+incompatible
-	github.com/container-storage-interface/spec v1.0.0
+	github.com/container-storage-interface/spec v1.3.0
 	github.com/go-logr/logr v0.2.1 // indirect
 	github.com/golang/protobuf v1.4.3
 	github.com/google/go-cmp v0.5.2
-	github.com/kubernetes-csi/csi-lib-utils v0.6.1
-	github.com/kubernetes-csi/csi-test v1.1.0
+	github.com/kubernetes-csi/csi-lib-utils v0.7.1
+	github.com/kubernetes-csi/csi-test v2.0.0+incompatible
```
tam7t, line 12: It looks like theres a v4.0.2 tag and the module named `github.com/kubernetes-csi/csi-test/v4`

`github.com/kubernetes-csi/csi-test/v4 v4.0.1` and referencing the packages as `github.com/kubernetes-csi/csi-test/v4` _may_ make this "compatible" again?

aramase: _checking_

test/sanity/sanity_test.go
aramase: This is for each individual sanity test, so we only need to cleanup the target path.

go.mod
aramase: Updated to the latest release of csi-test

tam7t (approved):
/lgtm

PR #394, opened 2020-11-24
feat: Reuse provider grpc clients

### LINKED ISSUE #300: Reusing GRPC client and connection for multiple mount events
Follow up for https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/280#discussion_r476702964

**Environment:**

tam7t:
Create the provider GRPC clients during process initialization instead
of on-demand. The Dial() call is non-blocking and if the connection is
disrupted grpc should re-establish the connection. By configuring RPCs
to retry through a service configuration, this will ensure that when a
plugin is restarted the connection will be re-dialed and the RPC
retried.

Fixes #300

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -29,8 +29,8 @@ const (
 	FailedToMount = "FailedToMount"
 	// SecretProviderClassNotFound error
 	SecretProviderClassNotFound = "SecretProviderClassNotFound"
-	// FailedToCreateProviderGRPCClient error
-	FailedToCreateProviderGRPCClient = "FailedToCreateProviderGRPCClient"
+	// FailedToLookupProviderGRPCClient error
+	FailedToLookupProviderGRPCClient = "FailedToLookupProviderGRPCClient"
 	// GRPCProviderError error
 	GRPCProviderError = "GRPCProviderError"
 	// FailedToRotate error
--- a/pkg/rotation/reconciler.go
+++ b/pkg/rotation/reconciler.go
@@ -83,7 +83,7 @@ type Reconciler struct {
 }
 
 // NewReconciler returns a new reconciler for rotation
-func NewReconciler(s *runtime.Scheme, providerVolumePath, nodeName string, rotationPollInterval time.Duration) (*Reconciler, error) {
+func NewReconciler(s *runtime.Scheme, providerVolumePath, nodeName string, rotationPollInterval time.Duration, providerClients map[string]*secretsstore.CSIProviderClient) (*Reconciler, error) {
 	config, err := buildConfig()
 	if err != nil {
 		return nil, err
@@ -109,7 +109,7 @@ func NewReconciler(s *runtime.Scheme, providerVolumePath, nodeName string, rotat
 		scheme:               s,
 		providerVolumePath:   providerVolumePath,
 		rotationPollInterval: rotationPollInterval,
-		providerClients:      make(map[string]*secretsstore.CSIProviderClient),
+		providerClients:      providerClients,
 		reporter:             newStatsReporter(),
 		queue:                workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
 		eventRecorder:        recorder,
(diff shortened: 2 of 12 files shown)


pkg/errors/errors.go
```diff
@@ -29,8 +29,8 @@ const (
 	FailedToMount = "FailedToMount"
 	// SecretProviderClassNotFound error
 	SecretProviderClassNotFound = "SecretProviderClassNotFound"
-	// FailedToCreateProviderGRPCClient error
-	FailedToCreateProviderGRPCClient = "FailedToCreateProviderGRPCClient"
+	// FailedToLookupProviderGRPCClient error
+	FailedToLookupProviderGRPCClient = "FailedToCreateProviderGRPCClient"
```
aramase, line 33:
```suggestion
	FailedToLookupProviderGRPCClient = "FailedToLookupProviderGRPCClient"
```

pkg/rotation/reconciler.go
```diff
@@ -272,11 +272,11 @@ func (r *Reconciler) reconcile(ctx context.Context, spcps *v1alpha1.SecretProvid
 	}
 
 	providerName = string(spc.Spec.Provider)
-	providerClient, err := r.getProviderClient(providerName)
-	if err != nil {
-		errorReason = internalerrors.FailedToCreateProviderGRPCClient
-		r.generateEvent(pod, v1.EventTypeWarning, mountRotationFailedReason, fmt.Sprintf("failed to create provider client, err: %+v", err))
-		return fmt.Errorf("failed to create provider client, err: %+v", err)
+	providerClient, exists := r.providerClients[providerName]
+	if !exists {
```
aramase, line 276: This logic is valid now as we support a mix of gRPC and old format for invoking the providers. The `--grpc-supported-providers` flag is the source of truth. When we make gRPC the default mode of communication between driver-provider, we'll need to have a fallback logic here to create the provider client on the fly is not found or we still need to continue setting the `--grpc-supported-providers` flag. WDYT?

pkg/secrets-store/secrets-store.go
```diff
@@ -94,12 +88,12 @@ func newIdentityServer(d *csicommon.CSIDriver) *identityServer {
...
 
 // Run starts the CSI plugin
-func (s *SecretsStore) Run(ctx context.Context, driverName, nodeID, endpoint, providerVolumePath, minProviderVersions, grpcSupportedProviders string, client client.Client) {
+func (s *SecretsStore) Run(ctx context.Context, driverName, nodeID, endpoint, providerVolumePath, minProviderVersions string, providerClients map[string]*CSIProviderClient, client client.Client) {
 	klog.Infof("Driver: %v ", driverName)
 	klog.Infof("Version: %s", vendorVersion)
 	klog.Infof("Provider Volume Path: %s", providerVolumePath)
 	klog.Infof("Minimum provider versions: %s", minProviderVersions)
-	klog.Infof("GRPC supported providers: %s", grpcSupportedProviders)
+	// klog.Infof("GRPC supported providers: %s", providerClients)
```
aramase, line 96: nit: Could we remove this line if not used? And can we log the `--grpc-supported-providers` flag in `main.go`. It'll be useful for debugging.

pkg/rotation/reconciler.go
tam7t: hmmm, good point, I need to think on that for a bit.

I'm generally a fan of explicit configuration like `--grpc-supported-providers` (I like a service's dependencies to not change during execution) but I'm also not a fan of having to edit the driver yaml for new/internal providers...

Another option would be to wrap the map in a structure with a mutex to create new clients on lookup misses. I think that would require a `--deprecated-exec-providers` flag to be added in the future.

ritazh: As a cluster operator, I can see benefits of setting the `--grpc-supported-providers` flag explicitly to control what driver-providers are allowed and rolled out in the clusters. 

But I can also see benefits of creating a provider client on the fly reduces driver downtime. 

Should we open an issue to discuss this more and get more user feedback on the desired experience?

aramase: I've opened an issue to discuss this: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/401.

This PR will change the current behavior where we create it on the fly. If we do go with this approach, then we should also add `azure` provider to the `--grpc-supported-providers` in all the manifests in the driver repo.

I'm not sure if users are actually piggybacking on `--grpc-supported-providers` to control which providers are used in the cluster. They would just want to configure the `SecretProviderClass` and assume the driver knows how to communicate with that provider which in the future by default will just be grpc.

aramase (commented):
/lgtm

I think we are good to go with this approach for the next release as we currently rely on `--grpc-supported-providers`. I've opened https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/401 for discussing creating provider clients on the fly.

PR #409, opened 2020-12-30
feat(values): add kubernetesRootDir

eyenx:

Adds `providersDir` to the `linux` and `windows` values to be able to configure the `hostPath` mountpoint of the `Daemonset`.

Not all Kubernetes Distributions set their `kubernetesRootDir` under `/etc/kubernetes` (example: Nutanix Karbon).

First PR in this repo.


ritazh:
Thanks for the PR @eyenx!
Can you please make these changes in the yamls in [manifest_staging/charts](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/master/manifest_staging/charts/secrets-store-csi-driver) where we host the staging charts? All the chart changes will then be promoted into the released charts folder with the next release. Thanks!

aramase (commented):
Thank you for the PR @eyenx.

1. As @ritazh pointed all the changes are required to be made in [manifest_staging/charts](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/master/manifest_staging/charts/secrets-store-csi-driver) folder.
2. I think it would be better to have the whole providers-dir configurable instead of just the kubernetes dir. So we can add `providerDir` and default is `/etc/kubernetes/secrets-store-csi-providers` for Linux and `C:\k\secrets-store-csi-providers` for Windows. I would love to hear what others think about this.
3. Please also add the new configurable values to the configuration [table](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/master/manifest_staging/charts/secrets-store-csi-driver#configuration)

eyenx:
@ritazh @aramase as requested, made my changes in the `manifest_staging` folder and switched to using `providersDir` instead of `kubernetesRootDir` and added the docs in the `README.md`

aramase (changes requested):
@eyenx Could you also update the values here:
- https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L55
- https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml#L53

eyenx:
Those lines aren't really corrupted by this change, as the `mountPath` is relevant inside the pod: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L103 And that can stay the same.

eyenx:
a.e. the registration Volume's `mountPath` is also hardcoded: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L38

aramase:
https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/418 already opened for `pull-secrets-store-csi-driver-e2e-gcp` failures.

tam7t:
I think this needs to become a requirement for supported providers actually. Currently the GCP provider install has hardcoded `providervol` set to `/etc/kubernetes/secrets-store-csi-providers`.

aramase:
> I think this needs to become a requirement for supported providers actually. Currently the GCP provider install has hardcoded providervol set to /etc/kubernetes/secrets-store-csi-providers.

It's the same for all providers. But I think we should add a note before next release that if this value is changed, then also make sure to update the volume in the provider manifests. WDYT?

I think with helm if the driver is packaged as part of provider, then it'll be handled.

tam7t:
@aramase ah that sounds good to me, realizing now that it is just a deploy configuration yaml parameter and not something that provider _code_ implementations really need to be aware of.

aramase (commented):
/lgtm

PR #413, opened 2021-01-04
helm: add config value for prometheus port

### LINKED ISSUE #411: Improvement about the port of driver metrics endpoint
I'm using Prometheus to gather the driver metrics. I thought parameter `metricsAddr` can control the port of the metrics endpoint. However, this endpoint only exposes controller related metrics. 

With the help of @mitsutaka, I realize that the port is [hard-coded](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/pkg/metrics/exporter.go#L26) as `8888`.  

I think there should be
1. Document about the port of the driver metrics. 
2. Parameter for overriding the port number. 

**Environment:**

- Secrets Store CSI Driver version: (use the image tag):  v0.0.18

aramase:
Add config value for prometheus port

Fixes #411 


manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
```diff
@@ -64,6 +64,9 @@ spec:
             - "--rotation-poll-interval={{ .Values.rotationPollInterval }}"
             {{- end }}
             - "--metrics-addr={{ .Values.windows.metricsAddr }}"
+            {{- if .Values.windows.prometheusPort }}
+            - "--prometheus-port={{ .Values.windows.prometheusPort }}"
```
ritazh, line 68: To be consistent, let's also explicitly add `--prometheus-port=` to the corresponding deployment yaml in /deploy/

aramase: Done!

mitsutaka:
@aramase What do you suppose to use `linux.metricsAddr` and `linux.prometheusPort` with Prometheus service discovery? Will the helm chart will cover annotation or depend on users scraping configurations?  Also, I'm wondering if these ports will be aggregated on one of the ports to expose both controller-runtime based metrics and CSI specific ones.

aramase:
> @aramase What do you suppose to use linux.metricsAddr and linux.prometheusPort with Prometheus service discovery? Will the helm chart will cover annotation or depend on users scraping configurations? Also, I'm wondering if these ports will be aggregated on one of the ports to expose both controller-runtime based metrics and CSI specific ones.

@mitsutaka I was exploring the same idea of combining the 2 ports but I think the handler path for the 2 metrics would need to be different. `controller-runtime` metrics get published on `/metrics`. Trying to use the same handler for driver metrics will yield no driver metrics on `/metrics`.

aramase:
ok, I found an elegant way to accomplish this. Take a look at https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/421.

Closing this in favor of https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/421

/close

mitsutaka:
@aramase awesome. Thanks for the PR

Issue #460, opened 2021-02-24
Proposal: secret-store-csi-driver file IO consolidation


tam7t:
**Describe the solution you'd like**
This is a tracking issue for the [secret-store-csi-driver file IO consolidation proposal](https://docs.google.com/document/d/1y1Oc1_-JGxMh-MZ57iqrVnoCIn6V-W0vJcj3qmq8EWU/edit#).

**Anything else you would like to add:**

**Environment:**

- Secrets Store CSI Driver version: (use the image tag):
- Kubernetes version: (use `kubectl version`):

PR #461, opened 2021-02-24
proto: add helpers and tools to install proto utilities

tam7t:

When modifying `provider/v1alpha1/service.proto` one must also regenerate the protobuf bindings for go.

This appears to have originally been done with:
* `protoc        v3.13.0`
* `protoc-gen-go v1.25.0`

But I was not actually able to find a released version of `protoc-gen-go` `v1.25.0`. The latest [grpc directions](https://grpc.io/docs/languages/go/quickstart/) instead show using the following:

```
go get google.golang.org/protobuf/cmd/protoc-gen-go \
         google.golang.org/grpc/cmd/protoc-gen-go-grpc
...
protoc --go_out=. --go_opt=paths=source_relative \
    --go-grpc_out=. --go-grpc_opt=paths=source_relative \
```

I've added a Makefile target and a dependency to `hack/tools` to be able to install `protoc v3.13.0` (prebuilt binary) and `protoc-gen-gov 1.23.0` from go mod.

Started as part of #460

I left comments to come back and get `protoc-gen-go` from `google.golang.org/protobuf/cmd/protoc-gen-go` instead of `github.com/golang/protobuf/protoc-gen-go` in the future, but don't want to change too many things at once.


Makefile
```diff
@@ -84,6 +86,7 @@ KIND_VERSION ?= 0.8.1
 KUBERNETES_VERSION ?= 1.18.2
 BATS_VERSION ?= 1.2.1
 TRIVY_VERSION ?= 0.14.0
+PROTOC_VERSION ?= 3.13.0
```
aramase, line 89: Why not use the latest version `v3.15.2`?

Makefile
```diff
@@ -330,7 +340,7 @@ manifests: $(CONTROLLER_GEN))
 
 .PHONY: generate-protobuf
 generate-protobuf: # generates protobuf
```
aramase, line 342: Can we also setup up the binaries here required for the target?

```suggestion
generate-protobuf: $(PROTOC) $(PROTOC_GEN_GO) # generates protobuf
```

Makefile
tam7t: I was bumbling my way through and wanted to not change too much at once. This looks like a simple update (no changes to generated code) so I'll bump to 3.15.2

Makefile
tam7t: ah nice!

aramase (commented):
/lgtm

aramase (commented):
/approve

PR #462, opened 2021-02-26
feat: connect to plugins at runtime instead of configuration

### LINKED ISSUE #401: Create provider grpc client dynamically
Currently the driver supports a mix (grpc or invoking binary) for communicating with the providers. With https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/394 the provider clients are created at start up based on providers in the `--grpc-supported-providers` list. If a new provider needs to be added, then the secrets-store-csi-driver daemonset manifest needs to be updated to add the new provider to the list.

Options:
1. The driver maintains a fallback logic similar to how its done today. When the provider client is not found, the client is created and then stored in the map on the fly.
2.  The driver only creates the provider clients during start up. If the user wishes to use a new provider, then they update the daemonset to add the new provider to list. This will cause the csi driver pods to restart and during this time any new pods on the node will need to wait to mount. 

**Environment:**

tam7t:

This creates an object that can be used to create and access plugin clients. Instead of building the connections at startup based on a flag of supported providers, an attempt to connect will be made on first use.

A regex is used to constrain the possible plugin paths that the driver will attempt to connect to.

This makes adding new plugins to a cluster easier since it will not require a new deployment of the driver.

Fixes #401


docs/book/src/providers.md
```diff
@@ -22,7 +22,7 @@ This document highlights the implementation steps for adding a secrets-store-csi
 
 ### Implementation details
 
-The driver as of `v0.0.14` adds an option to use gRPC to communicate with the provider. This is an alpha feature and is introduced with a feature flag `--grpc-supported-providers`. The `--grpc-supported-providers` is a `;` delimited list of all providers that support gRPC for communication. The driver will communicate with the provider using gRPC only if the provider name is in the list of supported providers in `--grpc-supported-providers`.
+The driver as of `v0.0.14` adds an option to use gRPC to communicate with the provider. This is an alpha feature and is introduced with a feature flag `--grpc-supported-providers`. The `--grpc-supported-providers` is a `;` delimited list of all providers that support gRPC for communication. This flag is no longer necessary as this is the only supported communication mechanism.
```
aramase, line 25: We should add a note about it being only supported mechanism as of v0.0.20 release of the driver

pkg/secrets-store/provider_client_test.go
```diff
@@ -200,3 +211,65 @@ func TestMountContentError(t *testing.T) {
...
+	var wg sync.WaitGroup
+
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		provider := fmt.Sprintf("server-%d", i)
+		go func() {
+			defer wg.Done()
+			if _, err := cb.Get(ctx, provider); err != nil {
+				t.Errorf("Get(%s) = %v, want nil", provider, err)
+			}
```
aramase, line 237: Can we also add a test for concurrent plugin get?

cmd/secrets-store-csi-driver/main.go
```diff
@@ -52,15 +51,12 @@ var (
 	logFormatJSON      = flag.Bool("log-format-json", false, "set log formatter to json")
 	providerVolumePath = flag.String("provider-volume", "/etc/kubernetes/secrets-store-csi-providers", "Volume path for provider")
 	// this will be removed in a future release
-	_           = flag.String("min-provider-version", "", "[DEPRECATED] set minimum supported provider versions with current driver")
-	metricsAddr = flag.String("metrics-addr", ":8095", "The address the metric endpoint binds to")
-	// grpcSupportedProviders is a ; separated string that can contain a list of providers. The reason it's a string is to allow scenarios
-	// where the driver is being used with 2 providers, one which supports grpc and other using binary for provider.
-	grpcSupportedProviders = flag.String("grpc-supported-providers", "", "set list of providers that support grpc for driver-provider [alpha]")
```
ritazh, line 59: As we have done with other flags, should we mark it as deprecated and to be removed, but not actually remove until 2 versions from now?

tam7t: Ah sounds good, I'll leave the flag in. I did intend to remove _setting_ the flag in the yaml and helm chart, so the upgrade path would immediately stop setting it, but I can see value in having the flag just be ignored for a few releases in case anyone is manually updating just the image tag in their yaml.

docs/book/src/providers.md
tam7t: Updated the note to say `v0.0.21` since the flag is still necessary in `v0.0.20` to know which providers to connect to.

pkg/secrets-store/provider_client_test.go
tam7t: Woops added TestPluginClientBuilder_ConcurrentGet

pkg/secrets-store/provider_client.go
```diff
@@ -20,63 +20,138 @@ import (
...
+				"RetryableStatusCodes": [ "UNAVAILABLE" ]
+			}
+		}
+	]
+}
+`
+
+var (
+	// PluginNameRe is the regular expression used to validate plugin names.
+	PluginNameRe        = regexp.MustCompile(`^[a-zA-Z0-9_-]{0,30}$`)
```
aramase, line 60: We should also add the provider name restrictions in the providers doc in a follow up PR

aramase (commented):
/lgtm

aramase (commented):
/lgtm

aramase:
@tam7t Sorry about the merge conflict 😢 

Could you rebase the PR? I'll them `lgtm` + `approve` this so we can get it in.

aramase (commented):
/lgtm
/approve

aramase:
The prow bot is a little slow today. It hasn't removed the `needs-rebase` label yet.

PR #482, opened 2021-03-18
test: update vault e2e to use versioned deployment

aramase:

This [PR](https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/pull/71) added livenessprobe and readiness probe for the vault provider manifests, however the current release still doesn't have the `/health` handler. So switching to using tagged manifests for vault provider e2e.

cc @tomhjp 

Fixes #

### DIFF
--- a/test/bats/vault.bats
+++ b/test/bats/vault.bats
@@ -6,7 +6,7 @@ BATS_TESTS_DIR=test/bats/tests/vault
 WAIT_TIME=120
 SLEEP_TIME=1
 NAMESPACE=default
-PROVIDER_YAML=https://raw.githubusercontent.com/hashicorp/secrets-store-csi-driver-provider-vault/master/deployment/provider-vault-installer.yaml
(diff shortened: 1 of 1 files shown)


aramase:
Created an issue in the vault provider for this: https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/issues/74. Once that issue is resolved we should switch back to using the latest manifests url.

aramase:
Ahh, looks like `0.0.7` isn't going to fix it either. The manifests tagged with that release are incorrect and were fixed in https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/pull/58. So we need the fix in master to update it.

aramase:
Opened https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/issues/74 to fix the issue.

aramase:
/close

As `0.0.7` release manifests are incorrect

PR #487, opened 2021-03-23
test: add make target for installing chart releases

aramase:

- Add Makefile target to deploy helm chart release. This will be used when we cut a release.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,11 +23,13 @@ REPO_PATH="$(ORG_PATH)/$(PROJECT_NAME)"
 
 REGISTRY ?= gcr.io/k8s-staging-csi-secrets-store
 IMAGE_NAME ?= driver
+# Release version is the current supported release for the driver
+# Update this version when the helm chart is being updated for release
+RELEASE_VERSION := v0.0.20
 IMAGE_VERSION ?= v0.0.21-rc.0
-E2E_IMAGE_VERSION = v0.1.0-e2e-$(BUILD_COMMIT)
 # Use a custom version for E2E tests if we are testing in CI
 ifdef CI
-override IMAGE_VERSION := $(E2E_IMAGE_VERSION)
+override IMAGE_VERSION := v0.1.0-e2e-$(BUILD_COMMIT)
 endif
 IMAGE_TAG=$(REGISTRY)/$(IMAGE_NAME):$(IMAGE_VERSION)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,8 +27,10 @@ IMAGE_VERSION ?= v0.0.21-rc.0
 E2E_IMAGE_VERSION = v0.1.0-e2e-$(BUILD_COMMIT)
 # Use a custom version for E2E tests if we are testing in CI
 ifdef CI
+ifndef RELEASE
```
tam7t, line 30: Is `RELEASE` referenced elsewhere here?

Could `e2e-helm-deploy` just point to `E2E_IMAGE_VERSION` instead of doing an override?

Should there be a `helm-deploy` target for non-e2e usecases?

Just trying to think of ways to reduce conditionals here as it can be kind of confusing to follow.

aramase:
Test-infra PR to add the release prow jobs: https://github.com/kubernetes/test-infra/pull/21489

Makefile
```diff
@@ -273,41 +275,46 @@ e2e-test: e2e-bootstrap # run test for windows
...
+		--set enableSecretRotation=true \
+		--set rotationPollInterval=30s
 
 .PHONY: e2e-kind-cleanup
 e2e-kind-cleanup:
 	kind delete cluster --name kind
 
 .PHONY: e2e-azure
-e2e-azure: $(AZURE_CLI) install-driver
+e2e-azure: $(AZURE_CLI) e2e-helm-deploy
```
aramase, line 309: We should remove `e2e-helm-deploy` from here after https://github.com/kubernetes/test-infra/pull/21489 is merged

Makefile
aramase: > Is RELEASE referenced elsewhere here?

Nope, it's only used for overriding the image version. 

> Could e2e-helm-deploy just point to E2E_IMAGE_VERSION instead of doing an override?

Yeah, as we're renaming the target to explicitly say `e2e-*`, I think it's safe to change the target to use `E2E_IMAGE_VERSION` instead of the override.

> Should there be a helm-deploy target for non-e2e usecases?

Typically users will just install it by adding the chart repo, but when we add the documentation for how to run e2e tests locally, we can add a target for developers who want to install the current driver. WDYT?

aramase: > Is RELEASE referenced elsewhere here?

Nope, it's only used for overriding the image version. 

> Could e2e-helm-deploy just point to E2E_IMAGE_VERSION instead of doing an override?

The `IMAGE_VERSION` is used in a couple of places other than installing the driver. For instance, image scanning and docker build for the image. We would need to replace all targets. So I think keeping just a single variable that's used across all targets makes it easier for developers trying to build the images locally too. WDYT?

> Should there be a helm-deploy target for non-e2e usecases?

Typically users will just install it by adding the chart repo, but when we add the documentation for how to run e2e tests locally, we can add a target for developers who want to install the current driver. WDYT?

Makefile
aramase: Now that test-infra PR is merged removed `e2e-helm-deploy` from the `e2e-*` targets.

Makefile
tam7t: I don't understand when `RELEASE` would be set is what is really confusing me here. I think a comment would help explain.

aramase: Had an offline discussion with Tommy and updated the PR to add a constant which makes it more clear:
```bash
RELEASE_VERSION := v0.0.21
IMAGE_VERSION ?= v0.0.21-rc.0
ifdef CI
override IMAGE_VERSION := v0.1.0-e2e-$(BUILD_COMMIT)
endif
```

tam7t: lgmt

PR #490, opened 2021-03-24
dial: grpc naming is not compatible with windows unix socket

tam7t:

Windows grpc connections are currently broken:

```
E0324 17:53:17.547284    6228 utils.go:79] GRPC error: failed to mount secrets store objects for pod default/nginx-secrets-store-inline-crd, err: rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = "transport: Error while dialing dial tcp: address unix://C:\\k\\secrets-store-csi-providers/azure.sock: too many colons in address"
```

Believe this was caused by https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/462 where `grpc.Dial` was replaced with [`grpc.DialContext`](https://pkg.go.dev/google.golang.org/grpc#DialContext).

Fixes #

### DIFF
--- a/pkg/secrets-store/provider_client.go
+++ b/pkg/secrets-store/provider_client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"regexp"
 	"sync"
@@ -109,10 +110,12 @@ func (p *PluginClientBuilder) Get(ctx context.Context, provider string) (v1alpha
 		return nil, fmt.Errorf("%w: provider %q", ErrProviderNotFound, provider)
 	}
 
-	conn, err := grpc.DialContext(
-		ctx,
-		fmt.Sprintf("unix://%s/%s.sock", p.socketPath, provider),
+	conn, err := grpc.Dial(
+		fmt.Sprintf("%s/%s.sock", p.socketPath, provider),
 		grpc.WithInsecure(), // the interface is only secured through filesystem ACLs
+		grpc.WithContextDialer(func(ctx context.Context, target string) (net.Conn, error) {
+			return (&net.Dialer{}).DialContext(ctx, "unix", target)
+		}),
 		grpc.WithDefaultServiceConfig(ServiceConfig),
 	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/secrets-store/provider_client.go
```diff
@@ -109,10 +110,15 @@ func (p *PluginClientBuilder) Get(ctx context.Context, provider string) (v1alpha
...
-		ctx,
-		fmt.Sprintf("unix://%s/%s.sock", p.socketPath, provider),
+	conn, err := grpc.Dial(
+		fmt.Sprintf("%s/%s.sock", p.socketPath, provider),
 		grpc.WithInsecure(), // the interface is only secured through filesystem ACLs
+		grpc.WithContextDialer(func(ctx context.Context, target string) (net.Conn, error) {
+			// context.Background() is used instead of the request context
+			// because we want the client to live and re-connect if needed
+			// beyond the request's context duration.
+			return (&net.Dialer{}).DialContext(context.Background(), "unix", target)
```
tam7t, line 120: Another option is to alter `NewPluginClientBuilder()` to accept the server-process context instead of using Background(), but since the main thread will call `Cleanup()` on interrupt i'm not sure how necessary that is.

tam7t: actually the ctx in this function is actually the dialer func, and grpc.Dial() will wrap Background() for us.

aramase:
/test pull-secrets-store-csi-driver-e2e-windows

rerunning because there were few flakes in the last run

aramase:
Looking at the kubelet logs, the failure seems to be windows related:
```bash
E0324 20:00:34.516993    4368 cni.go:364] Error adding default_nginx-secrets-store-inline-crd/7f0098aba9421d5a4fc9984d50db88905f1ee173acb296179156984faa73f423 to network azure-vnet/azure: A virtual machine or container with the specified identifier does not exist.
E0324 20:00:34.516993    4368 cni_windows.go:59] error while adding to cni network: A virtual machine or container with the specified identifier does not exist.
W0324 20:00:34.516993    4368 docker_sandbox.go:400] failed to read pod IP from plugin/docker: networkPlugin cni failed on the status hook for pod "nginx-secrets-store-inline-crd_default": A virtual machine or container with the specified identifier does not exist.
```

aramase (commented):
/lgtm
/approve

PR #493, opened 2021-03-24
fix: prevent duplicate owner references

aramase:

Came across this issue while testing with older Kubernetes releases where the API server allows duplication of owner references. Although it doesn't have an impact on the core functionality of gc, it could lead to large objects if the owner refs are duplicated. This was fixed in this PR: https://github.com/kubernetes/kubernetes/pull/96185. 

Fixing the check to ensure we don't append the same owner reference twice and updating tests to conform with it. 

Fixes #

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller_test.go
+++ b/controllers/secretproviderclasspodstatus_controller_test.go
@@ -172,7 +172,8 @@ func TestPatchSecretWithOwnerRef(t *testing.T) {
 	client := fake.NewFakeClientWithScheme(scheme, initObjects...)
 	reconciler := newReconciler(client, scheme, "node1")
 
-	err = reconciler.patchSecretWithOwnerRef(context.TODO(), "my-secret", "default", ref)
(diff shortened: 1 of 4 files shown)


controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -172,7 +172,7 @@ func TestPatchSecretWithOwnerRef(t *testing.T) {
 	client := fake.NewFakeClientWithScheme(scheme, initObjects...)
 	reconciler := newReconciler(client, scheme, "node1")
 
-	err = reconciler.patchSecretWithOwnerRef(context.TODO(), "my-secret", "default", ref)
+	err = reconciler.patchSecretWithOwnerRef(context.TODO(), "my-secret", "default", ref, ref)
```
tam7t, line 175: nit: a comment here explaining why `ref` is passed twice (or a separate test case for the double ref)

aramase: added a comment

Issue #463, opened 2021-02-27
Helm chart doesn't use kube-system namespace by default


andrebriggs:
**What steps did you take and what happened:**
Ran the latest Helm chart using Helm 3

```bash
$ helm template blah . > secret-store.yaml
$ cat secret-store.yaml
---
# Source: secrets-store-csi-driver/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: secrets-store-csi-driver
  namespace: default
  labels:
    app.kubernetes.io/instance: "blah"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/name: "secrets-store-csi-driver"
    app.kubernetes.io/version: "0.0.20"
    app: secrets-store-csi-driver
    helm.sh/chart: "secrets-store-csi-driver-0.0.20"
# (snip...)
```
**What did you expect to happen:**
Expected to see `ServiceAccount` object being the `kube-system` namespace based upon #375. From my understanding this is preferred for all objects that can be namespace according to #386 

I suggest making a note in the Helm chart [README.md](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/master/charts/secrets-store-csi-driver#configuration) recommending folks set the namespace to `kube-system` and/or add a variable in `values.yaml` to optionally set  `kube-system` to be the default namespace.

aramase:
Closed with #505 

/close

Issue #268, opened 2020-07-31
Update readme post next release


ritazh:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

**What did you expect to happen:**
We should add the following:
- default value for `syncSecrets.enable` will be set to `false` after the next 2 releases (secure by default)
- update deployment instructions https://github.com/kubernetes-sigs/secrets-store-csi-driver#optional-sync-with-kubernetes-secrets to enable this feature

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Which provider are you using:**
[e.g. Azure Key Vault, HashiCorp Vault, etc. Have you checked out the provider's repo for more help?]

**Environment:**

- Secrets Store CSI Driver version: (use the image tag):
- Kubernetes version: (use `kubectl version`):

aramase:
/assign @tam7t 

We should add this as best practices + bring it up in the next community call

PR #505, opened 2021-04-12
docs: kube-system, upgrades, best-practices

tam7t:

Adds additional documentation for addressing install and upgrade issues.

This addresses:

https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/463 - update helm install docs to set `kube-system`
https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/268 - add best practices doc for syncing rbac policy

Additionally consolidated breaking change notes to a new `upgrades` documentation page.

In https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/268 we said we would change the default after two releases, but multiple releases later and still have not changed the default. Proposing a lower friction step forward to adding a best-practices doc to cover the secret syncing feature.


aramase (commented):
@tam7t You would also need to add the new pages to https://raw.githubusercontent.com/kubernetes-sigs/secrets-store-csi-driver/master/docs/book/src/SUMMARY.md for the docs to show up.

docs/book/src/getting-started/upgrades.md
```diff
@@ -0,0 +1,38 @@
+# Upgrades
+
+This page includes instructions for upgrading the driver to the latest version.
+
+```bash
+helm upgrade csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver
```
ritazh, line 6:
```suggestion
helm upgrade csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver --namespace kube-system
```

docs/book/src/topics/best-practices.md
```diff
@@ -0,0 +1,30 @@
...
+
+If you need to use synchronise your external secrets to Kubernetes `Secret`s consider configuring encryption of data at
+rest: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/
+
+This will ensure that data is encrypted before it is stored in `etcd`.
+
+1. Keep the driver up to date
+
+Subscribe to the
+[`kubernetes-secrets-store-csi-driver`](https://groups.google.com/forum/#!forum/kubernetes-secrets-store-csi-driver)
```
ritazh, line 26: Can we also add "subscribe to github releases"?

ritazh:
> In #268 we said we would change the default after two releases, but multiple releases later and still have not changed the default. 

For changes to default values and deprecations, we could also consider adding the issue to a future release milestone to make sure we don't forget it.

docs/book/src/getting-started/upgrades.md
aramase: If the previous installation is in a different namespace, then this command will fail. We could use `NAMESPACE` var here and recommend replacing with the namespace where there current release is in.

docs/book/src/getting-started/upgrades.md
```diff
@@ -0,0 +1,38 @@
...
+
+## pre `v0.0.20`
+
+`v0.0.20` removed support for non-gRPC based providers. Follow your provider
+documentation to upgrade providers to use gRPC before upgrading the driver to
+`v0.0.20` or greater.
+
+## pre `v0.0.18`
+
+`v0.0.17` and earlier installed the driver to the `default` namespace.
```
aramase, line 20: Should we explicitly call out using yamls here instead of helm charts?

docs/book/src/topics/best-practices.md
```diff
@@ -0,0 +1,29 @@
+# Best Practices
```
tam7t, line 1: Should also have best practices for RBAC like not allowing users to edit `SecretProviderClassByPodStatus` CRDs

ritazh: I also like this section from the gcp provider: https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp#security-considerations section should we add them here as well?

docs/book/src/getting-started/upgrades.md
tam7t: Updated, let me know if this is what you meant by `NAMESPACE`

docs/book/src/topics/best-practices.md
tam7t: Added an item for this. Hopefully it provides the right context/call to action?

aramase (commented):
/lgtm

PR #508, opened 2021-04-13
feat: add optional healthcheck for provider plugins

### LINKED ISSUE #503: Add an optional healthcheck flag for providers
- Add an optional healthcheck flag for performing healthcheck with providers. We could use the version RPC request to check if the provider responds. The healthcheck will only be for provider clients that have been initialized.

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/58c586d3bb3286115ef0ec2fd44b4e5419334b9a/provider/v1alpha1/service.proto#L31-L43

**Environment:**

aramase:

/kind feature

- Adds an optional health check option for initialized provider plugin clients
  - `--provider-health-check=true` will enable health check for the plugin clients
  - `--provider-health-check-interval` used to control how often we want to perform the provider plugin health check
- The health check is done by making a `Version()` RPC call to the provider
- Failure to respond to health check will not result in any drastic action. We'll generate an error log that could be informative to users.

Fixes #503


aramase:
/retitle feat: add optional healthcheck for provider plugins

pkg/secrets-store/provider_client.go
```diff
@@ -154,6 +155,35 @@ func (p *PluginClientBuilder) Cleanup() {
...
+func (p *PluginClientBuilder) EnableHealthCheck(interval time.Duration, stopCh <-chan struct{}) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-stopCh:
+			return
+		case <-ticker.C:
+			p.lock.Lock()
```
ritazh, line 169: Maybe RLock is better here?

pkg/secrets-store/provider_client.go
```diff
@@ -154,6 +155,35 @@ func (p *PluginClientBuilder) Cleanup() {
...
+			defer p.lock.Unlock()
+
+			for provider, client := range p.clients {
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
+
+				runtimeVersion, err := Version(ctx, client)
+				if err != nil {
+					klog.V(4).ErrorS(err, "provider healthcheck failed", "provider", provider)
```
ritazh, line 178: would it be helpful to also log the err?

cmd/secrets-store-csi-driver/main.go
```diff
@@ -64,6 +64,10 @@ var (
 	// This feature flag will be enabled by default after n+2 releases giving time for users to label all their existing credential secrets.
 	filteredWatchSecret = flag.Bool("filtered-watch-secret", false, "enable filtered watch for NodePublishSecretRef secrets with label secrets-store.csi.k8s.io/used=true")
 
+	// Enable optional healthcheck for provider clients that exist in memory
+	providerHealthCheck         = flag.Bool("provider-health-check", false, "Enable health check for configured providers")
```
ritazh, line 68: Is this something we want to expose to the helm chart?

pkg/secrets-store/provider_client.go
aramase: Yeah, we are already logging the error here. Did you mean reducing the verbosity?

aramase:
Another option instead of using a custom health check logic would be to use https://github.com/grpc/grpc/blob/master/doc/health-checking.md. All providers would need to export that proto. The Azure provider already does. Then we can add the healthcheck func as part of `grpc.Dial()`

cmd/secrets-store-csi-driver/main.go
```diff
@@ -147,6 +151,12 @@ func main() {
 	providerClients := secretsstore.NewPluginClientBuilder(*providerVolumePath)
 	defer providerClients.Cleanup()
 
+	// enable provider health check
+	if *providerHealthCheck {
+		klog.InfoS("provider health check enabled", "interval", *providerHealthCheckInterval)
+		go providerClients.EnableHealthCheck(*providerHealthCheckInterval, ctx.Done())
```
tam7t, line 157: maybe just pass in the context as a parameter instead of `ctx.Done()` so that the context can also be used within `EnableHealthCheck` for each Version() rpc.

pkg/secrets-store/provider_client.go
ritazh: nvm i read it too fast

pkg/secrets-store/provider_client.go
```diff
@@ -154,6 +155,36 @@ func (p *PluginClientBuilder) Cleanup() {
 	p.conns = make(map[string]*grpc.ClientConn)
 }
 
+// EnableHealthCheck enables periodic healthcheck for configured provider clients by making
+// a Version() RPC call. If the provider healthcheck fails, we log an error.
+func (p *PluginClientBuilder) EnableHealthCheck(interval time.Duration, stopCh <-chan struct{}) {
```
tam7t, line 160: Since this actually performs the health check and blocks indefinitely, I think `HealthCheck` may be a better name.

It would also be helpful to describe the blocking nature in the function comment so that caller knows stopping conditions.

pkg/secrets-store/provider_client.go
```diff
@@ -154,6 +155,36 @@ func (p *PluginClientBuilder) Cleanup() {
...
+
+	for {
+		select {
+		case <-stopCh:
+			return
+		case <-ticker.C:
+			p.lock.RLock()
+
+			for provider, client := range p.clients {
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
tam7t, line 172: if the context is passed in then `ctx` can be used instead of `Background()`

aramase:
/hold

for reviews from @ritazh and @tam7t

pkg/secrets-store/provider_client.go
```diff
@@ -154,6 +155,36 @@ func (p *PluginClientBuilder) Cleanup() {
...
+				defer cancel()
+
+				runtimeVersion, err := Version(ctx, client)
```
tam7t, line 175: No need to change in this PR, but the service config defined above could be modified to add default request timeouts to the entire service or per-method:

https://github.com/grpc/grpc/blob/master/doc/service_config.md

pkg/secrets-store/provider_client_test.go
```diff
@@ -334,3 +334,42 @@ func TestPluginClientBuilderErrorInvalid(t *testing.T) {
 		t.Errorf("Get(%s) = %v, want %v", "bad/provider/name", err, ErrInvalidProvider)
 	}
 }
+
+func TestVersion(t *testing.T) {
```
tam7t, line 338: What do you think about a test case for `EnableHealthCheck` itself? It would be nice to at least have something that covers the `p.lock.RLock()` through `p.lock.RUnlock()`

cmd/secrets-store-csi-driver/main.go
aramase: The context in main is the context created with background + signal handlers. For the `HealthCheck` we should use a smaller context that terminates sooner for fail-fast instead of this context which will only cancel when pod is terminating. WDYT?

aramase: Unless we use the `ctx` from here as parent context for the `context.WithCancel` in healthcheck

tam7t: yeah basically by passing in the server context we can use that as the parent instead of `context.Background()` _and_ we'll pickup the cancellation signals, etc automatically

aramase: sgtm

pkg/secrets-store/provider_client_test.go
aramase: Done!

Issue #500, opened 2021-04-07
Atomic file updates


tam7t:
**Describe the solution you'd like**

The implementation of K8s Secrets uses symbolic links to present an atomic update of the entire mount to the application when secrets change (some K8s discussion of [secret updates](https://github.com/kubernetes/kubernetes/issues/18372)).

The [current implementation](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/481) in the driver and most (all?) providers updates a file at a time.

Creating this bug to track any observed issues with current behavior and whether we can switch to using `kubernetes/pkg/volume/util/atomic_writer.go` directly without modification.

**Anything else you would like to add:**

In the CSI Driver the targetPath (root of tmpfs) is of the form:

```
/var/lib/kubelet/pods/<pod id>/volumes/kubernetes.io~csi/<volume id>/mount
```

This path's contents are directly visible to the pod. In order to use the `atomic_writer` it would need to manage files in `../data_tmp` and other parent paths:

```
/var/lib/kubelet/pods/<pod id>/volumes/kubernetes.io~csi/<volume id>/data_tmp
```

Since the kubelet(?) populates `target_path` in `NodePublishVolumeRequest`, I'm unclear whether its possible to write to files in the parent directory. [ref](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodepublishvolume)

tam7t:
Also note that we currently create the tmpfs mount at the `target_path`, to use the `atomic_writer.go` the other sibling folders that atomic_writer uses to stage files should _also_ be tmpfs.

tam7t:
🤦‍♂️ 

I may have missread the atomic_writer docs, it seems that the `..data` paths are not in the parent directory but part of the filenames and visible to the pod itself. From the host:

```
$ sudo ls -la /var/lib/kubelet/pods/903a43b4-4553-4e05-acfa-669e842e1dde/volumes/kubernetes.io~secret/mypodserviceaccount-token-4xmpw
total 4
drwxrwxrwt 3 root root  140 Apr 20 19:24 .
drwxr-xr-x 3 root root 4096 Apr 20 19:24 ..
drwxr-xr-x 2 root root  100 Apr 20 19:24 ..2021_04_20_19_24_41.755804101
lrwxrwxrwx 1 root root   31 Apr 20 19:24 ..data -> ..2021_04_20_19_24_41.755804101
[...]

PR #520, opened 2021-04-28
feat: use the atomic file writer

tam7t:

When files are returned in the grpc response, the driver will use the `atomic_writer.go` code to write the files into the mount. This is the same file writer used for `ConfigMap` and `Secret` objects currently.

#481 only really included the file path validation logic. This PR updates the vendored code unchanged (other than package name and comment).

Benefits include:
* re-use k8s code
* supports nested paths
* will remove files that are not in the update

Unfortunately files written the "regular" way are not updated by the atomic writer, so I've added unit tests and a helper `cleanupProviderFiles` to ensure that a path mounted using the old style will be correctly updated if the new style is deployed.

Additionally using the atomic writer means that paths that start with `..` are not supported since the `atomic_writer` uses these as a scratch pad for making the updates before updating symlinks.

* #500

Since the atomic writer relies on symlinks, locations where `filepath.Walk` is used must be examined for correctness since that function does not follow symlinks.

### DIFF
--- a/pkg/secrets-store/nodeserver_test.go
+++ b/pkg/secrets-store/nodeserver_test.go
@@ -363,7 +363,7 @@ func TestNodeUnpublishVolume(t *testing.T) {
 			name: "Success for a mounted volume with a retry",
 			nodeUnpublishVolReq: csi.NodeUnpublishVolumeRequest{
 				VolumeId:   "testvolid1",
-				TargetPath: tmpdir.New(t, "", `*\\pods\\fakePod\\volumes\\kubernetes.io~csi\\myvol\\mount`),
+				TargetPath: tmpdir.New(t, "", `*mount`),
 			},
 			mountPoints:        []mount.MountPoint{},
 			shouldRetryUnmount: true,
--- a/pkg/secrets-store/provider_client_test.go
+++ b/pkg/secrets-store/provider_client_test.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"reflect"
 	"sync"
 	"testing"
@@ -33,6 +32,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/test_utils/tmpdir"
+	"sigs.k8s.io/secrets-store-csi-driver/pkg/util/fileutil"
 	"sigs.k8s.io/secrets-store-csi-driver/provider/fake"
 	"sigs.k8s.io/secrets-store-csi-driver/provider/v1alpha1"
 )
@@ -107,6 +107,33 @@ func TestMountContent(t *testing.T) {
 				"bar": 0777,
 			},
 		},
+		{
+			name:       "provider response with nested files",
+			permission: "777",
+			files: []*v1alpha1.File{
+				{
+					Path:     "foo",
+					Mode:     0644,
+					Contents: []byte("foo"),
+				},
+				{
+					Path:     "baz/bar",
+					Mode:     0777,
+					Contents: []byte("bar"),
+				},
+				{
+					Path:     "baz/qux",
+					Mode:     0777,
+					Contents: []byte("qux"),
+				},
+			},
+			objectVersions: map[string]string{"foo": "v1"},
+			expectedFiles: map[string]os.FileMode{
+				"foo":     0644,
+				"baz/bar": 0777,
+				"baz/qux": 0777,
+			},
+		},
 	}
 
 	for _, test := range cases {
@@ -140,18 +167,17 @@ func TestMountContent(t *testing.T) {
 
 			// check that file was written
 			gotFiles := make(map[string]os.FileMode)
-			filepath.Walk(targetPath, func(path string, info os.FileInfo, err error) error {
-				// skip mount folder
-				if path == targetPath {
-					return nil
-				}
-				rel, err := filepath.Rel(targetPath, path)
+			paths, err := fileutil.GetMountedFiles(targetPath)
+			if err != nil {
+				t.Fatalf("unable to read mounted files: %s", err)
+			}
+			for rel, abs := range paths {
+				info, err := os.Lstat(abs)
 				if err != nil {
-					return err
+					t.Fatalf("unable to read mounted files: %s", err)
 				}
 				gotFiles[rel] = info.Mode()
-				return nil
-			})
+			}
 
 			if diff := cmp.Diff(test.expectedFiles, gotFiles); diff != "" {
 				t.Errorf("MountContent() file mismatch (-want +got):\n%s", diff)
--- a/pkg/secrets-store/provider_client_test.go
+++ b/pkg/secrets-store/provider_client_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
@@ -64,6 +65,7 @@ func TestMountContent(t *testing.T) {
 		files          []*v1alpha1.File
 		// expectations
 		expectedFiles map[string]os.FileMode
+		skipon        string
 	}{
 		{
 			name:           "provider successful response (no files)",
@@ -77,13 +79,13 @@ func TestMountContent(t *testing.T) {
 			files: []*v1alpha1.File{
 				{
 					Path:     "foo",
-					Mode:     0644,
+					Mode:     0666,
 					Contents: []byte("foo"),
 				},
 			},
 			objectVersions: map[string]string{"foo": "v1"},
 			expectedFiles: map[string]os.FileMode{
-				"foo": 0644,
+				"foo": 0666,
 			},
 		},
 		{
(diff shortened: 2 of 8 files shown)


pkg/secrets-store/provider_client.go
```diff
@@ -230,6 +230,8 @@ func MountContent(ctx context.Context, client v1alpha1.CSIDriverProviderClient,
 		if err := fileutil.Validate(resp.GetFiles()); err != nil {
 			return nil, internalerrors.FileWriteError, err
 		}
+		// if files have been written by the driver, we want those cleaned up
+		// before WritePayloads
```
tam7t, line 234: delete, i made this part of WritePayloads

pkg/util/fileutil/writer_test.go
```diff
@@ -312,24 +272,153 @@ func TestWritePayloads(t *testing.T) {
...
+				},
+			},
+			second: []*v1alpha1.File{
+				{
+					Path:     "a/c/bar.txt",
+					Mode:     0644,
+					Contents: []byte("root"),
+				},
+			},
+			removed: []string{"a/foo.txt"},
```
tam7t, line 350: `a/b/foo.txt` should not exist.

Also a test case where multiple files are deleted

pkg/util/fileutil/writer.go
```diff
@@ -51,72 +35,58 @@ func Validate(payloads []*v1alpha1.File) error {
...
-		fullPath := filepath.Join(path, payload.Path)
-		baseDir, _ := filepath.Split(fullPath)
-
-		if err := os.MkdirAll(baseDir, os.ModePerm); err != nil {
-			return fmt.Errorf("unable to create directory %s: %w", baseDir, err)
-		}
+	// cleanup any payload paths that may have been written by a previous
+	// version of the driver/provider.
+	if err := cleanupProviderFiles(path, payloads); err != nil {
+		return err
```
tam7t, line 45: wrap this error to give more context that the issue occurred to the cleanup of existing files

pkg/secrets-store/provider_client_test.go
```diff
@@ -104,6 +104,33 @@ func TestMountContent(t *testing.T) {
...
+					Mode:     0644,
+					Contents: []byte("foo"),
+				},
+				{
+					Path:     "baz/bar",
+					Mode:     0777,
+					Contents: []byte("bar"),
+				},
+				{
+					Path:     "baz/qux",
```
tam7t, line 122: invalid path on windows

pkg/secrets-store/provider_client_test.go
```diff
@@ -104,6 +104,33 @@ func TestMountContent(t *testing.T) {
 				"bar": 0777,
```
tam7t, line 104: on windows:
```
            -   "bar": s"-rwxrwxrwx",
            +   "bar": s"-rw-rw-rw-",
            -   "foo": s"-rw-r--r--",
            +   "foo": s"-rw-rw-rw-",
```

pkg/util/fileutil/filesystem_test.go
```diff
@@ -64,7 +64,29 @@ func TestGetMountedFiles(t *testing.T) {
 				return dir
 			},
 			expectedErr: false,
-			expectedKey: "subdir/secret.txt",
+			want:        []string{"subdir/secret.txt"},
```
tam7t, line 67: on windows:

```
    --- FAIL: TestGetMountedFiles/target_path_dir/dir/file_found (0.00s)
        filesystem_test.go:106: GetMountedFiles() keys mismatch (-want +got):
              []string{
            -   "subdir/secret.txt",
            +   `subdir\secret.txt`,
              }
        tmpdir.go:39: error cleaning up tmp dir: remove C:\Users\tommy\AppData\Local\Temp\ut2093061619\subdir\secret.txt: The process cannot access the file because it is being used by another process.
```

pkg/util/fileutil/writer_test.go
```diff
@@ -226,71 +142,115 @@ func TestValidate_Error(t *testing.T) {
 func TestWritePayloads(t *testing.T) {
 	cases := []struct {
 		name    string
-		payload []*v1alpha1.File
+		first   []*v1alpha1.File
+		second  []*v1alpha1.File
+		removed []string
 	}{
 		{
 			name: "simple payload",
```
tam7t, line 150: all of these get wrong file modes on windows:

```
    --- FAIL: TestWritePayloads/simple_payload (0.01s)
        writer_test.go:364: WritePayload(first) could not be read: unexpected file mode. got: -rw-rw-rw-, want: -rw-r--r--
        writer_test.go:374: WritePayload(second) could not be read: unexpected file mode. got: -rw-rw-rw-, want: -rw-r--r--
```

pkg/util/fileutil/writer_test.go
```diff
@@ -226,71 +142,115 @@ func TestValidate_Error(t *testing.T) {
...
+			second: []*v1alpha1.File{
+				{
+					Path:     "foo",
+					Mode:     0644,
+					Contents: []byte("second"),
 				},
 			},
 		},
 		{
 			name: "simple payload - mode",
```
tam7t, line 167: on windows:

```
    --- FAIL: TestWritePayloads/simple_payload_-_mode (0.01s)
        writer_test.go:364: WritePayload(first) could not be read: unexpected file mode. got: -r--r--r--, want: -r--r-----
        writer_test.go:374: WritePayload(second) could not be read: unexpected file mode. got: -rw-rw-rw-, want: -rw-r--r--
```

pkg/util/fileutil/filesystem_test.go
tam7t: I think the `targetPath` func() for this test case needs to get the return values of os.Create and call `f.Close()`

pkg/secrets-store/provider_client_test.go
tam7t: I believe the file permissions diffs are due to:

```
// On Windows, only the 0200 bit (owner writable) of mode is used; it
// controls whether the file's read-only attribute is set or cleared.
// The other bits are currently unused. For compatibility with Go 1.12
// and earlier, use a non-zero mode. Use mode 0400 for a read-only
// file and 0600 for a readable+writable file.
```

https://golang.org/src/os/file.go?s=15847:15891#L522

pkg/util/fileutil/filesystem.go
```diff
@@ -29,37 +28,101 @@ var (
...
+			if err != nil {
+				continue
+			}
+
+			// the root path to walk
+			// i.e. targetPath/..data/foo
+			root := filepath.Join(targetPath, actual)
+
+			if err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+				// if there was an error walking path immediately propagate it
```
aramase, line 80: The majority of `filepath.Walk` in the if..else loop is similar except for the paths. Could we extract this into a common function?

tam7t: After sleeping on this I've refactored a few times to reduce duplication and hopefully it reads better now

pkg/secrets-store/nodeserver_test.go
```diff
@@ -363,7 +363,7 @@ func TestNodeUnpublishVolume(t *testing.T) {
 			name: "Success for a mounted volume with a retry",
 			nodeUnpublishVolReq: csi.NodeUnpublishVolumeRequest{
 				VolumeId:   "testvolid1",
-				TargetPath: tmpdir.New(t, "", `*\\pods\\fakePod\\volumes\\kubernetes.io~csi\\myvol\\mount`),
+				TargetPath: tmpdir.New(t, "", `*mount`),
```
tam7t, line 366: the path was not relevant to the test case and it error'd on windows because the `pattern` included path separator characters. on linux it passes because it treats the `\\` as part of the file name, not path separators.

pkg/util/fileutil/atomic_writer.go
```diff
@@ -0,0 +1,463 @@
...
+//    by the caller.
+//
+// The visible files in this volume are symlinks to files in the writer's data
+// directory.  Actual files are stored in a hidden timestamped directory which
+// is symlinked to by the data directory. The timestamped directory and
+// data directory symlink are created in the writer's target dir.  This scheme
+// allows the files to be atomically updated by changing the target of the
+// data directory symlink.
+//
+// Consumers of the target directory can monitor the ..data symlink using
```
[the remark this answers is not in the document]
tam7t: hm my understanding is that the inotify can work on the 'normal' path too since the symlink is updated on changes? Or maybe I'm misunderstanding how processes should watch for changes to normal K8s secrets?

tam7t:
/remove-hold

tests now pass on windows

pkg/util/fileutil/filesystem.go
```diff
@@ -29,37 +28,84 @@ var (
...
+	if err != nil {
+		return nil, err
 	}
-	for _, file := range files {
-		paths[file] = targetPath + sep + file
+
+	// for each file/directory in the targetPath, walk that entry
+	for _, entry := range d {
+		// skip the reserved paths of targetPath/..*
+		if strings.HasPrefix(entry.Name(), "..") {
```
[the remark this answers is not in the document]
tam7t: Yeah, this matches the `..2016_02_01_15_04_05.12345678` and `..data` reserved sub directories of `targetPath` that may exist on a subsequent mount request.

aramase (commented):
/approve

PR #526, opened 2021-05-04
cleanup: remove windows file deletion on unmount

tam7t:

the k8s.io/mount-utils package includes necessary logic:
https://github.com/kubernetes/mount-utils/blob/77a9b3dc97f9e87135fbdc74e84ab9611bcf9a1c/mount_windows.go#L215

I noticed this when going through some windows specific code paths and think its probably better to rely on the mounter to take care of this logic instead of including it in our unmount handler

N/A


aramase (commented):
/lgtm
/approve

PR #530, opened 2021-05-06
Helm 3 and CRDs

### LINKED ISSUE #286: Move crds to crds dir for helm3
- Move crds from `templates` to `crds` which is the recommended way for installing crds using helm3
- Add `template/crds.yaml` with 

  ```yaml
  {{- if .Values.installCRDs }}
  {{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
    {{ $.Files.Get $path }}
  ---
  {{- end }}
  {{- end }}
  ```
`installCRDs` by default will be false so crd install using helm2 is skipped. If using helm2 then `--set installCRDs=true`

**Environment:**

/good-first-issue

nlamirault:

Change CRD directory for Helm 3.

Fixes #286


aramase:
Hey @nlamirault. Thank you for the PR. There is an open PR to add this change which we have been holding off because the change to helm3 will be a breaking change for the helm charts (moving from template to `crds` dir would mean the crds get deleted during `helm upgrade`). I think the path forward will probably be to support helm3 and when the change goes in, recommend users to do a uninstall and then clean install.

Issue #541, opened 2021-05-13
Release management document


nilekhc:
Define release management and semantic versioning for project

Issue #457, opened 2021-02-19
Add upgrade to e2e


ritazh:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
A e2e job that tests upgrade from n to n+1 via helm

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Secrets Store CSI Driver version: (use the image tag):
- Kubernetes version: (use `kubectl version`):

tam7t:
1. install the release chart (version in `makefile`)
2. run the test suite
3. install the staging chart
4. run the test suite

nilekhc:
I'll take this.

PR #555, opened 2021-05-18
Release management

nilekhc:

Defines Release Management guidelines.

Fixes https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/541


docs/Release_Management.md
```diff
@@ -0,0 +1,50 @@
...
+- Secrets Store CSI Driver is currently tracking towards first stable release(v1.0.0) with [this](https://github.com/kubernetes-sigs/secrets-store-csi-driver/milestone/5) milestone.
+
+
+
+## Release Cadence and Branching
+- Secrets Store CSI Driver follows `monthly` release schedule.
+
+- A new release should be created on the _`second Wednesday`_ of each month. This schedule not only allows us to do bug fixes, but also provides an opportunity to address underline image vulnerabilities etc. if any.
+
+- Once the new version is decided(as per above guideline), release candidate branch should be created from `master` with name `vX.Y.Z-rc.W` followed by release with same tag, for eg. if new version is going to be `v0.1.2` then release candidate branch should be `v0.1.2-rc.0`
```
tam7t, line 41: there's some discrepancies between this and our current release docs:

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/docs/RELEASE.md#building-and-releasing

we use `release-<VERSION>` for the branch names (no rc's)

@aramase do you think this or the release docs should be updated? Should we also create long lived release branches?

docs/Release_Management.md
```diff
@@ -0,0 +1,50 @@
...
+## Versioning
+This project strictly follows [semantic versioning](https://semver.org/spec/v2.0.0.html). All releases will be of the form _vX.Y.Z_ where X is the major version, Y is the minor version and Z is the patch version.
+
+### Patch releases
+- Patch releases provide users with bug fixes and security fixes. They do not contain new features.
+
+### Minor releases
+- Minor releases contain security and bug fixes as well as _**new features**_.
+
+- They are backwards compatible.
```
tam7t, line 24: Are we committing to backwards compatible within the Major release? 

Does backwards compatible imply upgrade path or just features?

I'm not sure that we would want to commit to backwards compatibility until v1 Major release and within a Major release I'm not sure we want to support arbitrary upgrades.

example:

* Should `0.2.0` be backwards compatible in features with `0.1.0`
* Should a direct upgrade from `1.1.0` to `1.3.0` be possible or may you need to go through `1.1.0 -> 1.2.0 -> 1.3.0` (from the test gate section above i assume that the user would need to step through each minor release)

nilekhc: I think major release should imply breaking change. For eg if we go from `1.0.0` -> `2.0.0` then we indicate there is breaking change and not all features would work as expected.

Also, backward compatibility I am thinking in terms of features and not upgrade path. 
>Should 0.2.0 be backwards compatible in features with 0.1.0

This is correct

>Should a direct upgrade from 1.1.0 to 1.3.0 be possible or may you need to go through 1.1.0 -> 1.2.0 -> 1.3.0 (from the test gate section above i assume that the user would need to step through each minor release)

I think it should be possible to go from `1.1.0` -> `1.3.0`. I don't believe we have any such restriction today. (Please correct me if I am wrong here)

As for backward compatibility we can document since which version/release we can support. WDYT?

docs/Release_Management.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Major releases
+- Major releases contain breaking changes. Breaking changes refer to schema changes, flag changes, and behavior changes of Secrets Store CSI Driver that may require a clean install during upgrade and it may introduce changes that could break backward compatibility.
```
aramase, line 27: For flag changes, do we need a major release? Following along the Kubernetes release pattern, if we want to deprecate an existing flag, we could do that in a minor release itself after n+2 releases?

I agree on schema and/or behavior changes would warrant a major release because that could technically change the way users will need to configure.

docs/Release_Management.md
```diff
@@ -0,0 +1,50 @@
...
+unless project evolves in important new directions and such release is necessary.
+
+- Secrets Store CSI Driver is currently tracking towards first stable release(v1.0.0) with [this](https://github.com/kubernetes-sigs/secrets-store-csi-driver/milestone/5) milestone.
+
+
+
+## Release Cadence and Branching
+- Secrets Store CSI Driver follows `monthly` release schedule.
+
+- A new release should be created on the _`second Wednesday`_ of each month. This schedule not only allows us to do bug fixes, but also provides an opportunity to address underline image vulnerabilities etc. if any.
```
aramase, line 39: `should be created` could mean there will be a release every month on that day? we can maybe say "second week" instead to provide a buffer?

docs/Release_Management.md
aramase: I think we should create long live release branches as part of `v0.1.0`. This could be the 1st release that we can explore options to cherry pick bug fixes and cut patch releases. 

Note: we'll also need to track pipeline changes to handle these scenarios of building from release branches.

As for the release branches, I don't think we need release branches for rc. We cut a release branch for the intended release `release-v0.1.0` and rc images will be published from that release branch with tags. After rc image validation, we'll build a release image from the same branch. WDYT?

*This is how kubernetes does it today*

docs/Release_Management.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Once the new version is decided(as per above guideline), release candidate branch should be created from `master` with name `vX.Y.Z-rc.W` followed by release with same tag, for eg. if new version is going to be `v0.1.2` then release candidate branch should be `v0.1.2-rc.0`
+
+- Run tests to ensure stability. If issues/bugs are found, submit patches to the RC's release branch and create a new RC with the tag `vX.Y.Z-rc.W+1`. Apply those same patches to the `master` branch. Repeat until the release is suitably stable.
```
aramase, line 43: We should merge the fix to master and then cherry pick to the release branch (*similar to kubernetes*)

docs/Release_Management.md
nilekhc: Make sense. I'll update the doc

docs/Release_Management.md
nilekhc: Sounds good.

docs/Release_Management.md
tam7t: ACK, I'm not sure its feasible for us to maintain and test all permutations of up update paths, so maybe just adding some wording that upgrades are only _tested_ for consecutive version numbers

docs/Release_Management.md
tam7t: that sgtm, and also the merging to master + cherry picking to release branches

docs/Release_Management.md
nilekhc: @tam7t  / @aramase Do we do any upgrade test for driver?

docs/Release_Management.md
nilekhc: updated

docs/Release_Management.md
tam7t: Since you're working on #457 - yes we intend to so this sgtm

aramase (commented):
/approve

PR #569, opened 2021-05-28
Give aws secret and parameter names uuid

lasred:

Fixes #


test/bats/tests/aws/BasicTestMountSPC.yaml
```diff
@@ -12,18 +12,18 @@ spec:
...
           objectAlias: "ParameterStoreTest2"
           objectType: "ssmparameter"
-        - objectName: "ParameterStoreRotationTest"
+        - objectName: "ParameterStoreRotationTest-${UUID}"
           objectType: "ssmparameter"
-        - objectName: "SecretsManagerRotationTest"
+        - objectName: "SecretsManagerRotationTest-${UUID}"
           objectType: "secretsmanager"
-        - objectName: "SecretsManagerTest1"
+        - objectName: "SecretsManagerTest1-${UUID}"
```
aramase, line 24: Maybe just use `$SM_TEST_1_NAME` instead of a constant + variable? Similar changes for all the `objectName` in this file.

lasred: yup that would be cleaner

Makefile
```diff
@@ -98,6 +98,7 @@ SHELLCHECK_VER ?= v0.7.2
 BUILD_TIMESTAMP_W_SEC := $(shell date +%Y-%m-%d-%H-%M-%S)
 EKS_CLUSTER_NAME := integ-cluster-$(BUILD_TIMESTAMP_W_SEC)
 AWS_REGION := us-west-2
+AWS_TEST_UUID := $(shell openssl rand -hex 6)
```
aramase, line 101: Why don't we set this variable in the bats test suite? Similar to

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/8a9ccd19fee03495e4b43f99ac10198bbfdf948c/test/bats/azure.bats#L19-L21

I think we should avoid having all these defaults in the `Makefile` and keep them in the script file or bats test file instead.

lasred: I initially set this variable in the bats test suite  - https://github.com/lasred/secrets-store-csi-driver/blob/master/test/bats/aws.bats#L13 

but the issue is that with bats, this file gets sourced multiple times(in setup, and per every test) so I had conflicting UUIDs in setup_file where I created the secrets and test "deploy aws secretproviderclass crd" where the UUID was used to create the secret provider class

So I opted for the variable to be defined in the Makefile and passed into the bats script

test/bats/aws.bats
```diff
@@ -7,34 +7,46 @@ SLEEP_TIME=1
 PROVIDER_YAML=https://raw.githubusercontent.com/aws/secrets-store-csi-driver-provider-aws/main/deployment/aws-provider-installer.yaml
 NAMESPACE=kube-system
 POD_NAME=basic-test-mount
-export REGION=us-west-2
+export REGION=$REGION
```
aramase, line 10: nit:

How about `export REGION=${REGION:-us-west-2}`. By doing this we don't need to pass the `REGION` in the `e2e-aws` target.

test/scripts/initialize_eks_cluster.bash
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-EKS_CLUSTER_NAME=$1
-IMAGE_VERSION=$2
+AWS_REGION=$1
```
aramase, line 3: Same here too. Maybe declare a variable here with the default value being `us-west-2` with an option to override by the caller?

aramase (commented):
/lgtm
/approve

Thanks!

PR #571, opened 2021-05-28
Added GOPROXY propagation to container builds

### LINKED ISSUE #570: Plumb GOPROXY into build containers

Some corporate environments block access to `proxy.golang.org`. When building secrets-store-csi-driver in a container, setting `GOPROXY=direct` is not propagated, and even when it is, the absence of `git` causes builds to fail.

Not really.

**Environment:**

Corporate networks that sinkhole `proxy.golang.org` to `chalupa-dns-sinkhole.corp.example.com` only its not `example.com`

micahhausler:

Some corporate environments block access to `proxy.golang.org`. When building secrets-store-csi-driver in a container, setting `GOPROXY=direct` is not propagated, and even when it is, the absence of `git` causes builds to fail.

Fixes #570

You're awesome. Remember that.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,11 +1,13 @@
 ARG BASEIMAGE=us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:buster-v1.7.0
 
-FROM golang:1.16-alpine as builder
+FROM golang:1.16 as builder
 WORKDIR /go/src/sigs.k8s.io/secrets-store-csi-driver
 ADD . .
 ARG TARGETARCH
 ARG TARGETOS
 ARG LDFLAGS
+ARG GOPROXY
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,11 +1,14 @@
 ARG BASEIMAGE=us.gcr.io/k8s-artifacts-prod/build-image/debian-base-amd64:buster-v1.7.0
 
 FROM golang:1.16-alpine as builder
+RUN apk -U add git
```
aramase, line 4: I think we can just switch the builder image to `golang:1.16` instead of alpine. That way we don't have to install individual packages.

aramase (commented):
/lgtm

aramase (commented):
/approve

PR #587, opened 2021-06-09
feat: add rbac role and binding for rotation

aramase:

- Adds rbac clusterrole and binding for rotation reconciler to use with `nodePublishSecretRef`

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -412,6 +412,14 @@ manifests: $(CONTROLLER_GEN) $(KUSTOMIZE)
 	@sed -i '1s/^/{{ if .Values.syncSecret.enabled }}\n/gm; $$s/$$/\n{{ end }}/gm' manifest_staging/charts/secrets-store-csi-driver/templates/role-syncsecret.yaml
 	@sed -i '1s/^/{{ if .Values.syncSecret.enabled }}\n/gm; s/namespace: .*/namespace: {{ .Release.Namespace }}/gm; $$s/$$/\n{{ end }}/gm' manifest_staging/charts/secrets-store-csi-driver/templates/role-syncsecret_binding.yaml
 
+	# Generate rotation specific RBAC
+	$(CONTROLLER_GEN) rbac:roleName=secretproviderrotation-role paths="./pkg/rotation" output:dir=config/rbac-rotation
+	$(KUSTOMIZE) build config/rbac-rotation -o manifest_staging/deploy/rbac-secretproviderrotation.yaml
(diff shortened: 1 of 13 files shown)


config/rbac-rotation/role.yaml
```diff
@@ -0,0 +1,16 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
aramase, line 3: Creating a separate file similar to how we do it today for secret sync rbac roles

pkg/rotation/reconciler.go
```diff
@@ -84,6 +84,8 @@ type Reconciler struct {
 	crdClient            versioned.Interface
 }
 
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
```
aramase, line 87: These permissions are required for secret rotation + `nodePublishSecretRef`. Now that we're setting `syncSecret.enabled=false` if the user uses rotation + mount + `nodePublishSecretRef` without sync secret, the rotation will fail. Hence this additional role and binding.

Makefile
```diff
@@ -412,6 +412,16 @@ manifests: $(CONTROLLER_GEN) $(KUSTOMIZE)
...
 
+	# Generate rotation specific RBAC
+	$(CONTROLLER_GEN) rbac:roleName=secretproviderrotation-role paths="./pkg/rotation" output:dir=config/rbac-rotation
+	$(KUSTOMIZE) build config/rbac-rotation -o manifest_staging/deploy/rbac-secretproviderrotation.yaml
+	cp config/rbac-rotation/role.yaml manifest_staging/charts/secrets-store-csi-driver/templates/role-rotation.yaml
+	cp config/rbac-rotation/role_binding.yaml manifest_staging/charts/secrets-store-csi-driver/templates/role-rotation_binding.yaml
+	@sed -i '1s/^/{{ if .Values.enableSecretRotation }}\n/gm; $$s/$$/\n{{ end }}/gm' manifest_staging/charts/secrets-store-csi-driver/templates/role-rotation.yaml
+	@sed -i '1s/^/{{ if .Values.enableSecretRotation }}\n/gm; s/namespace: .*/namespace: {{ .Release.Namespace }}/gm; $$s/$$/\n{{ end }}/gm' manifest_staging/charts/secrets-store-csi-driver/templates/role-rotation_binding.yaml
+
+.PHONY: generate-protobuf
```
tam7t, line 423: is this a duplicate of line 425?

pkg/rotation/reconciler.go
aramase: Good point! Added a comment and TODO.

PR #545, opened 2021-05-17
fix: windows targetpath cleanup as part of node unpublish

aramase:
Signed-off-by: Anish Ramasekar <anish.ramasekar@gmail.com>

**What this PR does / why we need it**:
- Fixes error when the file contents still exist
```bash
I0517 18:47:49.349641    3820 utils.go:76] GRPC request: {"target_path":"c:\\var\\lib\\kubelet\\pods\\5a9c8959-98de-4a7d-b3a9-36ee6b25c9cb\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount","volume_id":"csi-fd831ea73cd95fa276590bbccaece8e5c0c4f03dd768f6585cbb6ac80f30dd12"}
W0517 18:47:49.349698    3820 mount_helper_common.go:129] Warning: "c:\\var\\lib\\kubelet\\pods\\5a9c8959-98de-4a7d-b3a9-36ee6b25c9cb\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount" is not a mountpoint, deleting
E0517 18:47:49.353861    3820 nodeserver.go:230] "failed to clean and unmount target path" err="remove c:\\var\\lib\\kubelet\\pods\\5a9c8959-98de-4a7d-b3a9-36ee6b25c9cb\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount: The directory is not empty." targetPath="c:\\var\\lib\\kubelet\\pods\\5a9c8959-98de-4a7d-b3a9-36ee6b25c9cb\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount"
```

With this fix included:
```bash
I0517 19:52:36.860648    2564 utils.go:76] GRPC request: {"target_path":"c:\\var\\lib\\kubelet\\pods\\750b1102-bc4d-4f53-a140-42df8868d5f7\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount","volume_id":"csi-aab4dde902888c5f309a35080999aef99d92bb9171784ed2ce9e8e497db87b05"}
W0517 19:52:36.862649    2564 mount_helper_common.go:129] Warning: "c:\\var\\lib\\kubelet\\pods\\750b1102-bc4d-4f53-a140-42df8868d5f7\\volumes\\kubernetes.io~csi\\secrets-store-inline\\mount" is not a mountpoint, deleting
[...]

PR #599, opened 2021-06-16
e2e: ensure pod deletion is successful

### LINKED ISSUE #546: integration tests for successful unmount
A test that ensures unmounts complete successfully on pod termination.

#526 removed some code and integration tests passed. It was only through manual verification that it was discovered that deletion before unmount was actually required: #545

**Environment:**

tam7t:

#526 removed some code and integration tests passed. It was only through manual verification that it was discovered that deletion before unmount was actually required: #545

Fixes #546

### DIFF
--- a/test/bats/aws.bats
+++ b/test/bats/aws.bats
@@ -147,10 +147,37 @@ teardown_file() {
     [[ "$result" == "SecretUser" ]]
 }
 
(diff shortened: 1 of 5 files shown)


tam7t:
Ah, i think if the pod is already deleted by the time the `kubectl wait` command is run we get `error: no matching resources found`, so will need to handle that case

aramase (commented):
/lgtm

PR #612, opened 2021-06-25
Possibility to annotate the created secret with CSI driver

### LINKED ISSUE #356: Possibility to annotate the created secret with CSI driver
Original issue - https://github.com/Azure/secrets-store-csi-driver-provider-azure/issues/279

At the moment the secrets can only be used within the namespace that the pod mounts the disk and secret is created. Using kubed we could share that secret with other namespaces. This will allow us to have single namespace with some alpine pod running all the mounts with secret and spread the secrets across namespaces.

For this to work the secret has to be annotated. At the moment it is not possible to annotate on creation/before. And every time you update the secret and redeploy mounts the manual annotation is removed.

I guess possibility of adding annotation to the secret on secreptorvider level would be enough.

**Environment:**

tetianakravchenko:
To support k8s secrets annotations

Fixes #356 

### DIFF
--- a/apis/v1alpha1/secretproviderclass_types.go
+++ b/apis/v1alpha1/secretproviderclass_types.go
@@ -47,8 +47,10 @@ type SecretObject struct {
 	// type of K8s secret object
 	Type string `json:"type,omitempty"`
 	// labels of K8s secret object
-	Labels map[string]string   `json:"labels,omitempty"`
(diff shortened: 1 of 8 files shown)


aramase (commented):
Thank you for the PR @tetianakravchenko!

/lgtm

PR #628, opened 2021-07-12
e2e: use kubectl wait to check if pods ready

aramase:

Only relies on `kubectl wait` to check if the pod is ready as part of the e2e. Using `wait_for_process` in addition to `kubectl wait` will lead the tests to timeout. This change will make the tests fail fast. 

Fixes #

### DIFF
--- a/test/bats/aws.bats
+++ b/test/bats/aws.bats
@@ -12,7 +12,6 @@ export REGION=${REGION:-us-west-2}
 export ACCOUNT_NUMBER=$(aws --region $REGION  sts get-caller-identity --query Account --output text)
 BATS_TEST_DIR=test/bats/tests/aws
 
-
 if [ -z "$UUID" ]; then 
    export UUID=secret-$(openssl rand -hex 6) 
 fi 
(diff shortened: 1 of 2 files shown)


test/bats/aws.bats
```diff
@@ -55,17 +54,15 @@ teardown_file() {
     run kubectl --namespace $NAMESPACE apply -f $PROVIDER_YAML  
     assert_success
 
-    cmd="kubectl --namespace $NAMESPACE wait --for=condition=Ready --timeout=60s pod -l app=csi-secrets-store-provider-aws"
-    wait_for_process $WAIT_TIME $SLEEP_TIME "$cmd"
+    kubectl --namespace $NAMESPACE wait --for=condition=Ready --timeout=120s pod -l app=csi-secrets-store-provider-aws
```
nilekhc, line 57: How to we arrive at timeout of 120s? I see 60s in some places. Is it based on past test runs?

aramase: The `120s` timeout is only for the provider image pulls because the image size isn't consistent for all providers. For the e2e cluster the image has to be pulled the 1st time and can take time. For the e2e test image, the image is small and the subsequent tests use `IfNotPresent` in the pod spec. So `60s` for the e2e pods is ample time for the provider to fetch content from external secrets store and driver to complete mount.

nilekhc (approved):
lgtm

PR #629, opened 2021-07-12
helm: move default annotations out of conditional

aramase:

Move the default container annotation out of the conditional.

Fixes #


manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
```diff
@@ -17,11 +17,11 @@ spec:
 {{ toYaml .Values.windows.updateStrategy | indent 4 }}
   template:
     metadata:
-    {{- if .Values.windows.podAnnotations }}
       annotations:
         kubectl.kubernetes.io/default-logs-container: secrets-store
```
aramase, line 21: Setting this annotation will use the `secrets-store` container by default for logs when `-c <container name>` isn't passed as part of `kubectl logs`

PR #656, opened 2021-07-23
feat: add keep-crd upgrade hook

aramase:

The crds have been moved from the templates/ to crds/ folder. When helm upgrade is run, helm will delete the crds because they're no longer in the generated template. To prevent deletion, we patch the 2 CRDs with the "helm.sh/resource-policy": "keep". Helm will skip deletion of resources with these annotation. Also, converted the hooks to a job as helm kills the pod before kubectl exec is run. With job, helm waits until the pod is run and reaches completion.

Fixes #

### DIFF
--- a/manifest_staging/charts/secrets-store-csi-driver/templates/keep-crds-upgrade-hook.yaml
+++ b/manifest_staging/charts/secrets-store-csi-driver/templates/keep-crds-upgrade-hook.yaml
@@ -1,71 +1,74 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ template "sscd.fullname" . }}-upgrade-crds
+  name: {{ template "sscd.fullname" . }}-keep-crds
 {{ include "sscd.labels" . | indent 2 }}
   annotations:
-    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook: pre-upgrade
     helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
-    helm.sh/hook-weight: "1"
+    helm.sh/hook-weight: "2"
 rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
-    verbs: ["get", "create", "update", "patch"]
+    verbs: ["get", "patch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "sscd.fullname" . }}-upgrade-crds
+  name: {{ template "sscd.fullname" . }}-keep-crds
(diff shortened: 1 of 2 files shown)


manifest_staging/charts/secrets-store-csi-driver/templates/crds-upgrade-hook.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
 {{ include "sscd.labels" . | indent 2 }}
   annotations:
     helm.sh/hook: pre-install,pre-upgrade
-    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+    helm.sh/hook-weight: "1"
```
aramase, line 9: The hooks are evaluated in ascending order based on the configured weights. We want the `keep-crds` hook to be the last one run during upgrade, so it annotates the CRDs.

manifest_staging/charts/secrets-store-csi-driver/templates/crds-upgrade-hook.yaml
```diff
@@ -36,31 +38,34 @@ metadata:
...
-  {{- if .Values.imagePullSecrets }}
-  imagePullSecrets:
-    {{ toYaml .Values.imagePullSecrets | indent 4 }}
-  {{- end }}
-  nodeSelector:
-    kubernetes.io/os: linux
+  backoffLimit: 0
+  template:
+    metadata:
+      name: crds-keep
```
nilekhc, line 58: Isn't this install/upgrade hook? Also can we uniform naming with `{{ template "sscd.fullname" . }}` prefix like other names?

manifest_staging/charts/secrets-store-csi-driver/templates/keep-crds-upgrade-hook.yaml
```diff
@@ -0,0 +1,74 @@
...
+{{ include "sscd.labels" . | indent 2 }}
+  annotations:
+    helm.sh/hook: pre-upgrade
+    helm.sh/hook-weight: "2"
+    helm.sh/hook-delete-policy: "hook-succeeded,before-hook-creation"
+spec:
+  backoffLimit: 0
+  template:
+    metadata:
+      name: crds-keep
```
nilekhc, line 58: same as above for prefixing with `{{ template "sscd.fullname" . }}`

manifest_staging/charts/secrets-store-csi-driver/templates/crds-upgrade-hook.yaml
aramase: this is only required during upgrade from the current charts `0.0.23` to newer versions. For fresh install, crds are installed by helm.

PR #671, opened 2021-07-29
helm: add option to configure fullnameOverride

### LINKED ISSUE #665: Truncate helm chart hook jobs name if longer than 63 char
Ref: https://github.com/Azure/secrets-store-csi-driver-provider-azure/issues/592

aramase:

Helm charts that package this as a dependency can override the fullname using the value. By default, helm appends the release chart name as prefix for the subchart which can make the name really long. Helm charts that package this as a dependency can use `fullnameOverride: secrets-store-csi-driver` to ensure the parent chart release name isn't appended to the driver fullname. 

Fixes #665 

### DIFF
--- a/manifest_staging/charts/secrets-store-csi-driver/templates/_helpers.tpl
+++ b/manifest_staging/charts/secrets-store-csi-driver/templates/_helpers.tpl
@@ -12,11 +12,15 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 If release name contains chart name it will be used as a full name.
 */}}
 {{- define "sscd.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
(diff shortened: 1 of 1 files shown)


manifest_staging/charts/secrets-store-csi-driver/templates/_helpers.tpl
```diff
@@ -12,13 +12,17 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 If release name contains chart name it will be used as a full name.
 */}}
 {{- define "sscd.fullname" -}}
+{{- if .Values.fullnameOverride -}}
```
nilekhc, line 15: Can we add this in doc for default value?

aramase: Did you mean adding the value to the helm configuration table?

`nameOverride` and `fullnameOverride` are already in the [configuration table](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/master/manifest_staging/charts/secrets-store-csi-driver#configuration). The default value is empty because we don't explicitly set anything in the `values.yaml`.

PR #682, opened 2021-08-10
test: implement e2e provider

### LINKED ISSUE #472: Consider an e2e-test provider
In working on #471, I realized it would be really nice to have a dummy provider that was only used for e2e tests in addition to 
the existing e2e tests:

This would be useful for several reasons:
* Testing end-to-end behavior that doesn't need any infrastructure set up (no GCP/Azure/AWS account, no vault installation) 
* Testing eventual API version upgrades from [sigs.k8s.io/secrets-store-csi-driver/provider/v1alpha1](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/provider/v1alpha1/service.proto)
* Verifying plugin behavior when [`RequiresRepublish`](https://kubernetes-csi.github.io/docs/csi-driver-object.html?highlight=RequiresRepublish#what-fields-does-the-csidriver-object-have) is set to `true`

Would this be a contribution the maintainers would like to see?

nilekhc:
This PR implements mock provider for e2e testing, thus eliminating dependency on cloud-provider specific Secret Store CSI driver provider. 

Fixes #472 


test/e2eprovider/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.16 as builder
+WORKDIR /go/src/sigs.k8s.io/secrets-store-csi-driver
+ADD . .
+RUN make build-e2e-provider
+RUN ls -latr test/e2eprovider
```
aramase, line 5: Was this added for debug?

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,144 @@
+package e2eprovider
```
aramase, line 1: We should add build tags to these files `// +build e2e` and update go build to build with the e2e tag

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+// Server is a mock csi-provider server
+type Server struct {
+	grpcServer *grpc.Server
+	listener   net.Listener
+	SocketPath string
+	network    string
+	Vault      vault.Vault
+}
+
+// NewE2eProviderServer returns a mock csi-provider grpc server
```
aramase, line 29: Let's standardize using mock/fake across all the changes here

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+// Start starts the mock csi-provider server
+func (m *Server) Start() error {
+	var err error
+
+	m.listener, err = net.Listen(m.network, m.SocketPath)
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("Listening for connections on address: %v", m.listener.Addr())
```
aramase, line 58: Let's use structured logging across all the files to be consistent

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, fmt.Errorf("objects is not set")
+	}
+
+	var objects types.StringArray
+	err = yaml.Unmarshal([]byte(objectsStrings), &objects)
+	if err != nil {
+		return nil, fmt.Errorf("failed to yaml unmarshal objects, error: %w", err)
+	}
+
+	keyVaultObjects := []types.KeyVaultObject{}
```
aramase, line 102: IMO we should rename these references to `externalSecretsStore` instead of key vault

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+			return nil, fmt.Errorf("unmarshal failed for keyVaultObjects at index %d, error: %w", i, err)
+		}
+
+		keyVaultObjects = append(keyVaultObjects, keyVaultObject)
+	}
+
+	for _, keyVaultObject := range keyVaultObjects {
+		fileName := keyVaultObject.ObjectName
+		if keyVaultObject.ObjectAlias != "" {
+			fileName = keyVaultObject.ObjectAlias
```
aramase, line 116: These are specific to azure provider and I don't think we need it here. We should have a simple logic for the RPC calls without too many validations as it's more to test the driver logic than provider. WDYT?

test/e2eprovider/server_test.go
```diff
@@ -1,99 +0,0 @@
...
-import (
-	"context"
-	"encoding/json"
-	"reflect"
-	"testing"
-
-	"sigs.k8s.io/secrets-store-csi-driver/provider/v1alpha1"
-)
-
-func TestMount(t *testing.T) {
```
aramase, line 12: It'll be good to have basic unit tests for the `e2e-provider`

test/e2eprovider/types/types.go
```diff
@@ -0,0 +1,24 @@
...
+	// the filename the object will be written to
+	ObjectAlias string `json:"objectAlias" yaml:"objectAlias"`
+	// the version of the Azure Key Vault objects
+	ObjectVersion string `json:"objectVersion" yaml:"objectVersion"`
+	// the type of the Azure Key Vault objects
+	ObjectType string `json:"objectType" yaml:"objectType"`
+	// the format of the Azure Key Vault objects
+	// supported formats are PEM, PFX
+	ObjectFormat string `json:"objectFormat" yaml:"objectFormat"`
+	// The encoding of the object in KeyVault
```
aramase, line 16: These are very specific to a single provider. We should simplify this to just name, version and return a static value that the driver can deterministically validate.

test/e2eprovider/vault/vault.go
```diff
@@ -0,0 +1,177 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+// E2eVault is a fake Vault for end-to-end testing.
+type E2eVault struct {
```
aramase, line 15: Let's make it more generic and call it `E2EStore`. WDYT?

test/e2eprovider/vault/vault.go
```diff
@@ -0,0 +1,177 @@
...
+	secret, err := v.KubeClient.CoreV1().Secrets(v.VaultNamespaceName).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return "", "", err
+	}
+
+	return string(secret.Data[name]), secret.GetLabels()["version"], nil
+}
+
+// populateVault populates the fake Vault with default secrets.
+func populateVault(kubeClient *kubernetes.Clientset) (string, error) {
```
aramase, line 51: IMO, this additional logic to create a backend with Kubernetes secret introduces an additional error path in the e2e which don't want to have. What if we just had the `e2e-provider` implementation trimmed down to return static values to perform a deterministic value check in the bats suite? That reduces the scope of failures to be specific for driver during e2e runs which is the intent. I like how we have are defining the interface to show how this can be useful for actual provider implementation but the `GetSecret` can just return a simple value?

cc @tam7t

pkg/secrets-store/nodeserver_test.go
```diff
@@ -18,14 +18,15 @@ package secretsstore
...
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"sigs.k8s.io/secrets-store-csi-driver/apis/v1alpha1"
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/secrets-store/mocks"
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/test_utils/tmpdir"
-	"sigs.k8s.io/secrets-store-csi-driver/test/e2eprovider"
+	server "sigs.k8s.io/secrets-store-csi-driver/test/e2eprovider/server"
```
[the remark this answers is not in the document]
nilekhc: It is required [here](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/682/files#diff-6d438949f97677c98d673fc22e1a3c43e2f1c4cc7a14e4973555395b8c9c1acbR278).

pkg/secrets-store/nodeserver_test.go
```diff
@@ -274,8 +275,9 @@ func TestNodePublishVolume(t *testing.T) {
 			r := mocks.NewFakeReporter()
 
 			tmpDir := tmpdir.New(t, "", "ut")
-			server, err := e2eprovider.NewSimpleCSIProviderServer(
-				filepath.Join(tmpDir, "simple_provider.sock"),
+			server, err := server.NewE2eProviderServer(
+				fmt.Sprintf("%s%s", "unix://", filepath.Join(tmpDir, "simple_provider.sock")),
```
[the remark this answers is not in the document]
nilekhc: done

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,363 @@
...
+export KEY_VERSION=${KEYVAULT_KEY_VERSION:-"v1"}
+export KEY_VALUE_CONTAINS=${KEYVAULT_KEY_VALUE:-"LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KVGhpcyBpcyBmYWtlIGtleQotLS0tLUVORCBQVUJMSUMgS0VZLS0tLS0K"}
+
+# export node selector var
+export NODE_SELECTOR_OS=$NODE_SELECTOR_OS
+
+# auto rotate secret name
+if [ -z "$AUTO_ROTATE_SECRET_NAME" ]; then
+    export AUTO_ROTATE_SECRET_NAME=secret-$(openssl rand -hex 6)
+fi
```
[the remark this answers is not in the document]
nilekhc: The variable names are used in tests. We don't need to set them, but can use their default value for testing.

test/e2eprovider/Dockerfile
nilekhc: yes.. removed.

test/e2eprovider/server/server.go
nilekhc: We need this in order to test `alias`.

pkg/secrets-store/nodeserver_test.go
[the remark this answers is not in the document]
aramase: The import is required but the package alias `server` isn't required as the package is already named `*/server`.

test/e2eprovider/server/server.go
nilekhc: Removed alias

pkg/secrets-store/nodeserver_test.go
```diff
@@ -274,8 +275,9 @@ func TestNodePublishVolume(t *testing.T) {
 			r := mocks.NewFakeReporter()
 
 			tmpDir := tmpdir.New(t, "", "ut")
-			server, err := e2eprovider.NewSimpleCSIProviderServer(
-				filepath.Join(tmpDir, "simple_provider.sock"),
+			server, err := server.NewE2eProviderServer(
```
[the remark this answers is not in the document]
nilekhc: done

pkg/secrets-store/nodeserver_test.go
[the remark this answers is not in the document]
nilekhc: I think it's because go modules.. The module name is `e2eprovider` but it's inside `server` directory.. so it explicitly needs the name

test/e2eprovider/server/server.go
nilekhc: updated to `mock`

test/e2eprovider/server/server.go
nilekhc: changed to `MockSecretsStoreObject`

test/e2eprovider/types/types.go
nilekhc: done

test/e2eprovider/vault/vault.go
nilekhc: removed vault

test/e2eprovider/vault/vault.go
nilekhc: switched back to returning `static values` return rather than `K8s secrets as vault`

test/e2eprovider/server/server.go
nilekhc: done

test/e2eprovider/server/server.go
nilekhc: done

pkg/secrets-store/nodeserver_test.go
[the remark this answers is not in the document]
tam7t: My preference would be to try and remove the dependency on e2eprovider from these tests, although the circular dependency seems to be cut by using the `replace`, i'm worried that having that will make issues for the other providers when we try to depend on `secrets-store-csi-driver`

test/e2eprovider/server_test.go
nilekhc: added tests

pkg/secrets-store/nodeserver_test.go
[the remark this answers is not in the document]
nilekhc: @tam7t  should we remove tests involving mock e2e provider?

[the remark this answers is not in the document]
nilekhc: Actually I just tried to trace and we are not doing anything with mock provider server in these test except initializing and testing for error.
For this we can rely on test written for e2r provider. I'll remove it.

cc @aramase @tam7t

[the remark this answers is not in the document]
nilekhc: I removed that dependency  and added test in mocker `server_test.go`

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+	grpcServer *grpc.Server
+	listener   net.Listener
+	SocketPath string
+	network    string
+}
+
+// NewE2EProviderServer returns a mock csi-provider grpc server
+func NewE2EProviderServer(endpoint string) (*Server, error) {
+	network := ""
+	address := ""
```
aramase, line 45:
```suggestion
	var network, address string
```

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+// Start starts the mock csi-provider server
+func (m *Server) Start() error {
+	var err error
+
+	m.listener, err = net.Listen(m.network, m.SocketPath)
+	if err != nil {
+		return err
+	}
+
+	klog.InfoS("Listening for connections on address:", m.listener.Addr())
```
aramase, line 77:
```suggestion
	klog.InfoS("Listening for connections", "address", m.listener.Addr())
```
this is using structured logging

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+	return resp, nil
+}
+
+func getSecret(secretName, podUID string) (*v1alpha1.File, *v1alpha1.ObjectVersion, error) {
+	secretVersion := "v1"
+	secretContent := secrets[secretName]
+
+	// If pod found in cache, then it means that pod is being called for the second time for rotation
+	// In this case, we should return the 'rotated' secret.
+	if ok := podCache[podUID]; ok {
```
aramase, line 152: If there are concurrent test runs, we can end up with concurrent map read/write. We should use a `RWMutex` for reading/writing in the cache

test/e2eprovider/server/server_test.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func TestMain(m *testing.M) {
+	setup()
+	exitCode := m.Run()
+	os.Exit(exitCode)
+}
+
+func setup() {
+	tempDir, err := ioutil.TempDir("", "")
```
aramase, line 28: `io/ioutil` is deprecated in go 1.16; use `os` instead

test/e2eprovider/server/server_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		panic(err)
```
aramase, line 30: pass `t *testing.T` as arg to the function and use `t.Error` instead

test/bats/tests/e2e_provider/e2e_provider_v1alpha1_secretproviderclass_ns.yaml
```diff
@@ -0,0 +1,49 @@
...
+    type: Opaque
+    data: 
+    - objectName: $SECRET_NAME
+      key: username
+  parameters:
+    objects: |
+      array:
+        - |
+          objectName: $SECRET_NAME
+          objectType: secret
```
aramase, line 44: `objectType` has no significance in the mock provider, so we should remove it in all the yamls

Makefile
```diff
@@ -228,7 +233,7 @@ test-style: lint lint-charts shellcheck
 .PHONY: lint
 lint: $(GOLANGCI_LINT)
 	# Setting timeout to 5m as default is 1m
-	$(GOLANGCI_LINT) run --timeout=5m -v
+	$(GOLANGCI_LINT) run --build-tags e2e --timeout=5m -v
```
aramase, line 236: `golangci-lint` will not work with nested modules, so we need to have another step to run `golangci-lint` inside the `test/e2eprovider` directory.

```bash
	cd test/e2eprovider && $(GOLANGCI_LINT) run --timeout=5m -v
```

thanks to @tam7t for the pointer

go.mod
```diff
@@ -14,14 +14,19 @@ require (
...
+	k8s.io/klog/v2 v2.10.0
 	k8s.io/mount-utils v0.21.0
 	sigs.k8s.io/controller-runtime v0.9.0
-	sigs.k8s.io/yaml v1.2.0
+	sigs.k8s.io/secrets-store-csi-driver/test/e2eprovider v0.0.0-00010101000000-000000000000
+)
+
+replace (
+	sigs.k8s.io/secrets-store-csi-driver => ./
+	sigs.k8s.io/secrets-store-csi-driver/test/e2eprovider => ./test/e2eprovider
```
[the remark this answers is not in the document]
nilekhc: removed

test/scripts/e2e_provider.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+echo "e2e provider test"
+make e2e-bootstrap e2e-mock-provider-container e2e-helm-deploy e2e-provider-deploy e2e-provider
```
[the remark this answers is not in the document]
nilekhc: done

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+
+var (
+	secrets = map[string]string{
+		"foo": "secret",
+		"fookey": `-----BEGIN PUBLIC KEY-----
+This is mock key
+-----END PUBLIC KEY-----`,
+	}
+
+	podCache = make(map[string]bool)
```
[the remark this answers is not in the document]
nilekhc: done

test/e2eprovider/e2e_provider.go
```diff
@@ -0,0 +1,47 @@
...
+	defer klog.Flush()
+
+	flag.Parse()
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT, os.Interrupt)
+
+	mockProviderServer, err := server.NewE2EProviderServer(*endpoint)
+	if err != nil {
+		klog.Fatalf("failed to get new fake e2e provider server, err: %+v", err)
```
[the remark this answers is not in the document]
nilekhc: done

test/e2eprovider/e2e_provider.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	if err := os.Remove(mockProviderServer.SocketPath); err != nil && !os.IsNotExist(err) {
+		klog.Fatalf("failed to remove %s, error: %s", mockProviderServer.SocketPath, err.Error())
+	}
+
+	err = mockProviderServer.Start()
+	if err != nil {
+		klog.Fatalf("failed to start fake e2e provider server, err: %+v", err)
```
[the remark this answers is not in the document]
nilekhc: done

test/e2eprovider/go.mod
```diff
@@ -0,0 +1,15 @@
...
+require (
+	google.golang.org/grpc v1.39.0
+	k8s.io/klog/v2 v2.10.0
+	sigs.k8s.io/secrets-store-csi-driver v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/yaml v1.2.0
+)
+
+replace (
+	sigs.k8s.io/secrets-store-csi-driver => ../..
+	sigs.k8s.io/secrets-store-csi-driver/test/e2eprovider => ./
```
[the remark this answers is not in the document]
nilekhc: `sigs.k8s.io/secrets-store-csi-driver` is required. Other one is removed

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+
+	podUIDAttribute = "csi.storage.k8s.io/pod.uid"
+)
+
+// Server is a mock csi-provider server
+type Server struct {
+	grpcServer *grpc.Server
+	listener   net.Listener
+	SocketPath string
```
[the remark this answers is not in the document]
nilekhc: updated

test/e2eprovider/server/server.go
nilekhc: done

test/e2eprovider/server/server.go
nilekhc: done

test/e2eprovider/server/server_test.go
nilekhc: done

test/e2eprovider/server/server_test.go
nilekhc: We do not have access to `testing.T` in `TestMain` method

Makefile
nilekhc: added

test/e2eprovider/server/server.go
nilekhc: done

test/bats/tests/e2e_provider/e2e_provider_v1alpha1_secretproviderclass_ns.yaml
nilekhc: removed

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	podCache = make(map[string]bool)
+
+	podUIDAttribute = "csi.storage.k8s.io/pod.uid"
+)
+
+// Server is a mock csi-provider server
+type Server struct {
+	grpcServer *grpc.Server
```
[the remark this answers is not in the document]
nilekhc: Composition is resulting into null pointer. Hence keeping this implementation

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return secretFile, version, nil
+}
+
+// Version implements provider csi-provider method
+func (s *Server) Version(ctx context.Context, req *v1alpha1.VersionRequest) (*v1alpha1.VersionResponse, error) {
+	return &v1alpha1.VersionResponse{
+		Version:        "v1alpha1",
+		RuntimeName:    "SimpleProvider",
```
aramase, line 185:
```suggestion
		RuntimeName:    "E2EMockProvider",
```

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,188 @@
...
+		RuntimeVersion: "0.0.10",
```
aramase, line 186:
```suggestion
		RuntimeVersion: "v0.0.10",
```

test/e2eprovider/server/server_test.go
```diff
@@ -0,0 +1,141 @@
...
+	testMockServer.Stop()
+	os.RemoveAll(tempDir)
+}
+
+func TestMockServer(t *testing.T) {
+	cases := []struct {
+		name string
+	}{
+		{
+			name: "start",
```
aramase, line 47: IMO we don't need table tests here as it's just checking the server start.

test/e2eprovider/server/server_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func TestMount(t *testing.T) {
+	cases := []struct {
+		name    string
+		input   *v1alpha1.MountRequest
+		want    *v1alpha1.MountResponse
+		wantErr error
+	}{
+		{
+			"Parse static secrets",
```
aramase, line 69: Maybe also add a test here for validating the rotation logic?

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,353 @@
...
+  sleep 60
+
+  result=$(kubectl exec -n rotation secrets-store-inline-rotation -- cat /mnt/secrets-store/$SECRET_NAME)
+  [[ "${result//$'\r'}" == "rotated" ]]
+
+  result=$(kubectl get secret -n rotation rotationsecret -o jsonpath="{.data.username}" | base64 -d)
+  [[ "${result//$'\r'}" == "rotated" ]]
+}
+
+@test "Test filtered-watch-secret=false" {
```
aramase, line 314: We can remove this test as filtered watch secret is enabled by default and `--filtered-watch-secret` is deprecated.

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,353 @@
...
+
+load helpers
+
+BATS_TESTS_DIR=test/bats/tests/e2e_provider
+WAIT_TIME=60
+SLEEP_TIME=1
+NODE_SELECTOR_OS=linux
+
+# export secret vars
+export SECRET_NAME=${KEYVAULT_SECRET_NAME:-foo}
```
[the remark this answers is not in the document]
nilekhc: done

test/bats/e2e-provider.bats
nilekhc: done

test/e2eprovider/server/server_test.go
nilekhc: Yes.. But I thought to be consistent with other tests. Also easier to read test result after running test on local or in pipeline. WDYT?

test/e2eprovider/server/server.go
nilekhc: done

test/e2eprovider/server/server.go
nilekhc: done

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,353 @@
...
+}
+
+teardown_file() {
+  #cleanup
+  run kubectl delete namespace non-filtered-watch
+  run kubectl delete namespace rotation
+  run kubectl delete namespace test-ns
+
+  run kubectl delete pods secrets-store-inline-crd secrets-store-inline-multiple-crd --force --grace-period 0
+}
```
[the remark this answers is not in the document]
nilekhc: done

test/e2eprovider/e2e-provider-installer.yaml
```diff
@@ -0,0 +1,87 @@
...
+kind: ClusterRole
+metadata:
+  name: csi-secrets-store-e2e-provider
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "update", "patch", "delete"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
+    verbs: ["get", "create", "update", "patch", "list", "delete"]
```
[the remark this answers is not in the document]
nilekhc: yup. removed

test/e2eprovider/e2e-provider-installer.yaml
```diff
@@ -0,0 +1,87 @@
...
+  selector:
+    matchLabels:
+      app: csi-secrets-store-e2e-provider
+  template:
+    metadata:
+      labels:
+        app: csi-secrets-store-e2e-provider
+    spec:
+      serviceAccountName: csi-secrets-store-e2e-provider
+      hostNetwork: true
```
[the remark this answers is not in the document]
nilekhc: not necessarily. removed

test/e2eprovider/server/server_test.go
nilekhc: done

aramase (commented):
@nilekhc I think the aws test failures for this PR are legit. The sleep for rotation is 40s in aws tests:

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/6fd1d310da5e2059ccc3f35af06debafc3ebc4e8/test/bats/aws.bats#L110-L111

and seen in the log, the rotation isn't yet complete:
```bash
not ok 6 CSI inline volume test with rotation - parameter store
# (in test file test/bats/aws.bats, line 113)
#   `[[ "${result//$'\r'}" == "AfterRotation" ]]' failed
# {
#     "Version": 2,
#     "Tier": "Standard"
# }
```
Ref: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_secrets-store-csi-driver/682/pull-secrets-store-csi-driver-e2e-aws/1432447020353720320

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,346 @@
...
+
+@test "CSI inline volume test with pod portability - read key from pod" {
+  result=$(kubectl exec secrets-store-inline-crd -- cat /mnt/secrets-store/$KEY_NAME)
+  result_base64_encoded=$(echo "${result//$'\r'}" | base64 ${BASE64_FLAGS})
+  [[ "${result_base64_encoded}" == *"${KEY_VALUE_CONTAINS}"* ]]
+}
+
+@test "CSI inline volume test with pod portability - unmount succeeds" {
+  skip
```
aramase, line 93: Looks like we're skipping all the tests after `CSI inline volume test with pod portability - read key from pod` in the suite?

test/bats/e2e-provider.bats
```diff
@@ -0,0 +1,346 @@
...
+
+  # On Linux a failure to unmount the tmpfs will block the pod from being
+  # deleted.
+  run wait $WAIT_PID
+  assert_success
+
+  # Sleep to allow time for logs to propagate.
+  sleep 10
+  # save debug information to archive in case of failure
+  archive_info
```
aramase, line 111: This test has been simplified in this PR: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/713. Could you update it to match the latest?

Makefile
```diff
@@ -404,7 +428,8 @@ e2e-helm-deploy:
 		--set linux.enabled=true \
 		--set syncSecret.enabled=true \
 		--set enableSecretRotation=true \
-		--set rotationPollInterval=30s
+		# Updated for mock provider testing. Mock provider returns default secret value as "rotated" once reconciler kicks in after 30s, but some e2e-provider tests takes more than 30s to complete before testing for rotation.
```
aramase, line 431: Could you provide more info on this behavior? This is the flow:

1. `Mount` RPC call from driver to provider as part of `NodePublishVolume`. At this point, the pod isn't in the mock provider cache. So it'll add the pod to the cache and return "secret".
2. After 30s, rotation reconciler calls make a `Mount` RPC call to the provider. Now, the pod is already in the cache, so return "rotated".

So, why would this increase in timeout be required? With this change we're also increasing the other provider tests to wait for `150s` which is long?

nilekhc: Step 2 intermittently occurs before we could run tests where we expect `secret` but gets `rotated`. Hence longer rotationPollInterval time to delay the rotation.

Initially I thought about increasing poll interval just for `e2e-provider` tests but implementation seems convoluted at the time given our make targets. But let me give it another try.

aramase: I'm fine making this change in a follow-up PR. We can get this PR merged after the other 2 comments are resolved.

test/bats/e2e-provider.bats
nilekhc: done

test/bats/e2e-provider.bats
nilekhc: done

Makefile
nilekhc: Actually I reverted the change of polling interval and implemented more deterministic behavior for rotation test

test/e2eprovider/e2e_provider.go
```diff
@@ -0,0 +1,60 @@
...
+		klog.Fatalf("failed to remove %s, error: %s", mockProviderServer.GetSocketPath(), err.Error())
+	}
+
+	err = mockProviderServer.Start()
+	if err != nil {
+		klog.Fatalf("failed to start mock e2e provider server, err: %+v", err)
+	}
+
+	// endpoint to enable rotation response.
+	// rotation response ("rotated") might be triggered by rotaion reconciler before other tests. This results in failure of those tests.
```
aramase, line 46:
```suggestion
	// rotation response ("rotated") might be triggered by rotation reconciler before other tests. This results in failure of those tests.
```

test/e2eprovider/server/server.go
```diff
@@ -0,0 +1,200 @@
...
+
+func getSecret(secretName, podUID string) (*v1alpha1.File, *v1alpha1.ObjectVersion, error) {
+	secretVersion := "v1"
+	secretContent := secrets[secretName]
+
+	// If pod found in cache, then it means that pod is being called for the second time for rotation
+	// In this case, we should return the 'rotated' secret.
+	m.RLock()
+	if ok := podCache[podUID]; ok {
+		if os.Getenv("ROTATION_ENABLED") == "true" { // ROTAION_ENABLED is set to true only when rotation tests are running
```
aramase, line 166: nit: move comment to next line

aramase (commented):
Just 2 small nits!

test/e2eprovider/server/server.go
nilekhc: done

test/e2eprovider/e2e_provider.go
nilekhc: done

aramase (commented):
/lgtm
/approve

PR #692, opened 2021-08-13
helm: adding annotations to upgrade jobs

### LINKED ISSUE #675: Job annotations
We use istio on our estate, which creates a sidecar to encrypt pod-pod communication.
This causes jobs to never complete because the side car never exits.
Istio can be turned off either by adding annotations to the pod or can be told to ignore pods with specific labels.
However we because both the daemonset and the job have the same labels, and you can't specify job annotations we can't tell istio to ignore the jobs keep-crds and upgrade-crds

Happy to create a PR if you are ok with it, I think the easiest is to extend the work done https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/435 to include annotations for the jobs
That way you could pass the value linux.jobAnnotations to annotate the 2 jobs
can also split it out so there is a keepjob and updatejob if you wanted to have the option of different annotations for the 2 jobs.

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L25
https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/charts/secrets-store-csi-driver/templates/keep-crds-upgrade-hook.yaml#L49
All the labels in keep-crds-upgrade-hook are also in csi-driver and there are no annotations for the job either

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): v0.1.0
- Kubernetes version: (use `kubectl version`): v1.20.5

thomasmRavn:

Fixes #675 

https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/675


thomasmRavn:
I signed it

thomasmRavn:
I signed it

aramase (commented):
/ok-to-test

aramase:
/label tide/merge-method-squash
/retitle helm: adding annotations to upgrade jobs

aramase (commented):
@thomasmRavn Thank you for the PR!

/approve

Issue #597, opened 2021-06-16
Deprecate `--filtered-watch-secret` flag


aramase:
/kind cleanup

sozercan:
Is there a deprecation process defined?

aramase:
This has been the process so far:

1. Deprecate the flag in the code (Setting the flag no longer takes effect)
2. Remove the flag completely after n+2 releases

616b2f:
@aramase any insides about why it is depricated? I mean it was set to true in v0.1.0 via this PR #550 and was mentioned in the release notes of the last release? Can you please help me to understand.

aramase:
> @aramase any insides about why it is depricated? I mean it was set to true in v0.1.0 via this PR #550 and was mentioned in the release notes of the last release? Can you please help me to understand.

The feature is enabled in `v0.1.0`. The flag will be deprecated as this is going to be default behavior and will not be configurable going forward.

PR #697, opened 2021-08-19
chore: make `--filtered-watch-secret` flag no-op

aramase:

Filtered watch secret has been enabled by default in `v0.1.0` release. This PR makes the flag configuration no-op to not allow disabling the feature. The flag will be removed altogether from code, deployment and helm charts in n+2 releases.

ref https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/597

### DIFF
--- a/cmd/secrets-store-csi-driver/main.go
+++ b/cmd/secrets-store-csi-driver/main.go
@@ -64,6 +64,7 @@ var (
 
 	// enable filtered watch for NodePublishSecretRef secrets. The filtering is done on the csi driver label: secrets-store.csi.k8s.io/used=true
 	// For Kubernetes secrets used to provide credentials for use with the CSI driver, set the label by running: kubectl label secret secrets-store-creds secrets-store.csi.k8s.io/used=true
+	// This feature is enabled by default starting v0.1.0 and can't be disabled starting v1.0.0 release.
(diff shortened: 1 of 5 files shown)


cmd/secrets-store-csi-driver/main.go
```diff
@@ -64,7 +64,8 @@ var (
 
 	// enable filtered watch for NodePublishSecretRef secrets. The filtering is done on the csi driver label: secrets-store.csi.k8s.io/used=true
 	// For Kubernetes secrets used to provide credentials for use with the CSI driver, set the label by running: kubectl label secret secrets-store-creds secrets-store.csi.k8s.io/used=true
-	filteredWatchSecret = flag.Bool("filtered-watch-secret", true, "enable filtered watch for NodePublishSecretRef secrets with label secrets-store.csi.k8s.io/used=true")
+	// This feature is enabled by default starting v0.1.0 and can't be disabled starting v1.0.0 release.
+	_ = flag.Bool("filtered-watch-secret", true, "enable filtered watch for NodePublishSecretRef secrets with label secrets-store.csi.k8s.io/used=true")
```
tam7t, line 68: should we add a `klog.Error` or warning if this is set to `false` to tell users that it has no effect?

aramase: Added a warning message: `Filtered watch for nodePublishSecretRef secret based on secrets-store.csi.k8s.io/used=true label can't be disabled. The --filtered-watch-secret flag will be deprecated in future releases.`. Lmk what you think.

PR #704, opened 2021-08-20
Updated metrics-addr to accept integer value rather than ':port-number'

Atharva-Shinde:
Signed-off-by: Atharva Shinde <atharvashinde179@gmail.com>

Fixes #


cmd/secrets-store-csi-driver/main.go
```diff
@@ -55,7 +55,7 @@ var (
 	logFormatJSON      = flag.Bool("log-format-json", false, "set log formatter to json")
 	providerVolumePath = flag.String("provider-volume", "/etc/kubernetes/secrets-store-csi-providers", "Volume path for provider")
 	// this will be removed in a future release
-	metricsAddr          = flag.String("metrics-addr", ":8095", "The address the metric endpoint binds to")
+	metricsAddr          = flag.String("metrics-addr", "8095", "The address the metric endpoint binds to")
```
tam7t, line 58: We should not have a configuration named `addr` when it actually accepts a port number as this will cause confusion.

cmd/secrets-store-csi-driver/main.go
```diff
@@ -112,7 +112,7 @@ func main() {
 
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme:             scheme,
-		MetricsBindAddress: *metricsAddr,
+		MetricsBindAddress: fmt.Sprintf(":%d", *metricsAddr),
```
tam7t, line 115: this loses the ability to choose how the metrics service should be bound (whether just local interface or all interfaces)

nilekhc: @tam7t Could you please elaborate?

nilekhc:
Hello all,
We discussed this PR today in our [community call](https://www.youtube.com/watch?v=99cvC_5K2IQ). We agreed that specifying address rather than a port number, gives user flexibility around what bind address they want to use and we want to persist that flexibility. We could potentially create new flag just to accept _portNumber_ but then it will increase confusion rather than simplifying things which was original motive behind this PR. So, we are not moving ahead with this change.

/close

PR #729, opened 2021-09-07
chore: update the initialDelaySeconds and timeoutSeconds for node-driver-registrar livenessprobe

aramase:

Sets the `initialDelaySeconds` and `timeoutSeconds` for the node-driver-registrar livenessprobe check. 
- The default `timeoutSeconds` of 1s is honored starting from Kubernetes 1.20 and is aggressive. Setting the value to 15s which is appropriate for linux and windows.
- Setting `initialDelaySeconds` to 30s to provide time for the socket to be created and driver to be registered during startup of the driver pods.
- Set resource requests to 100m/100Mi for all the containers. The limits are already 200m/200Mi for the `node-driver-registrar`, `livenessprobe` containers and 400m/400Mi for the secrets-store container.

Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes for defaults

Fixes #

### DIFF
--- a/manifest_staging/deploy/secrets-store-csi-driver.yaml
+++ b/manifest_staging/deploy/secrets-store-csi-driver.yaml
@@ -28,7 +28,8 @@ spec:
               - /csi-node-driver-registrar
               - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-secrets-store/csi.sock
               - --mode=kubelet-registration-probe
-            initialDelaySeconds: 3
+            initialDelaySeconds: 30
+            timeoutSeconds: 15
           env:
             - name: KUBE_NODE_NAME
               valueFrom:
--- a/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
+++ b/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
@@ -47,7 +47,8 @@ spec:
               - /csi-node-driver-registrar.exe
               - --kubelet-registration-path={{ .Values.windows.kubeletRootDir }}\plugins\csi-secrets-store\csi.sock
               - --mode=kubelet-registration-probe
-            initialDelaySeconds: 3
+            initialDelaySeconds: 30
+            timeoutSeconds: 15
           env:
           - name: KUBE_NODE_NAME
             valueFrom:
--- a/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
+++ b/manifest_staging/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
@@ -47,7 +47,8 @@ spec:
               - /csi-node-driver-registrar
               - --kubelet-registration-path={{ .Values.linux.kubeletRootDir }}/plugins/csi-secrets-store/csi.sock
               - --mode=kubelet-registration-probe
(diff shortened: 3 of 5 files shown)


nilekhc:
Could you comment on how we came up with these new values?

aramase:
> Could you comment on how we came up with these new values?

There are no recommended values for the liveness probe. In case of windows this can take a while before the the socket is created and is available. The reason for the PR is to override the current defaults which are aggressive and can cause restarts. There is more info on the defaults in the PR description.

aramase:
/hold

/cc @marosset 
for thoughts on default for windows

manifest_staging/deploy/secrets-store-csi-driver.yaml
```diff
@@ -28,7 +28,8 @@ spec:
               - /csi-node-driver-registrar
               - --kubelet-registration-path=/var/lib/kubelet/plugins/csi-secrets-store/csi.sock
               - --mode=kubelet-registration-probe
-            initialDelaySeconds: 3
+            initialDelaySeconds: 30
+            timeoutSeconds: 15
```
marosset, line 32: I cannot comment on the resources section but I think that having container limits set to 100m / 100Mi is too little resources for an exec probe to behave deterministically.

@dcantah @kevpar Do we have a rough idea around how much resources are needed to start a new process in a running container? These limits seem too low to accommodate the process set as the container entry point and have something else run periodically.

gedlu:
i  am waiting for this fix. My the CSI windows driver is complaining and the error is liveness probe failed. And also it is intermittent. This fix probably the one we need.

aramase:
@gedlu The primary issue is being investigated here: https://github.com/kubernetes/kubernetes/issues/104584. The changes to timeout and resource will be done as part of this PR and included in the next release.

manifest_staging/deploy/secrets-store-csi-driver.yaml
aramase: @dcantah @kevpar any thoughts on @marosset's comment?

dcantah: Shouldn't be any more expensive than the same process(es) launched directly on the host. Regardless of that, all of the system services and the init process/any additional execs in the container should be counting towards this 100MB limit set. It might be a bit aggressive depending on what the registrar does. A simple container that just runs `ping -t 127.0.0.1`  consumes about 32-35MB on my machine for example.

aramase: @marosset Given the `node-driver-registrar` is only in charge for registering the driver should we use 100m/100Mi instead of the current defaults? I know we were using 200m/200Mi for our repro testing but the behavior wasn't deterministic even with that.

marosset: I'm OK with using 100m/100Mi in this case. (especially with the longer `initialDelaySeconds` for the exec probe) and we can adjust if necessary in the future.

aramase: sounds good! I've updated the PR to set requests to 100m/100Mi for all the containers. The limits are already 200m/200Mi for the `node-driver-registrar`, `livenessprobe` containers and 400m/400Mi for the `secrets-store` container. PTAL when you get a chance!

PR #751, opened 2021-09-24
docs: add supported kubernetes versions

aramase:

- Adds supported kubernetes versions for the driver

Fixes #

This change was inspired by the supported kubernetes version note in [gatekeeper](https://github.com/open-policy-agent/gatekeeper)

### DIFF
--- a/docs/Release_Management.md
+++ b/docs/Release_Management.md
@@ -44,6 +44,18 @@ This project strictly follows [semantic versioning](https://semver.org/spec/v2.0
 
 - Any `fixes` or `patches` should be merged to main and then `cherry pick` to the release branch.
 
+## Supported Releases
+
+Applicable fixes, including security fixes, may be cherry-picked into the release branch, depending on severity and feasibility. Patch releases are cut from that branch as needed.
+
+We expect users to stay reasonably up-to-date with the versions of Secrets Store CSI Driver they use in production, but understand that it may take time to upgrade. We expect users to be running approximately the latest patch release of a given minor release and encourage users to upgrade as soon as possible.
+
+We expect to "support" n (current) and n-1 major.minor releases. "Support" means we expect users to be running that version in production. For example, when v1.3.0 comes out, v1.1.x will no longer be supported for patches and we encourage users to upgrade to a supported version as soon as possible.
+
+## Supported Kubernetes Versions
+
+Secrets Store CSI Driver will maintain support for all actively supported Kubernetes minor releases per [Kubernetes Supported Versions policy](https://kubernetes.io/releases/version-skew-policy/). If you choose to use Secrets Store CSI Driver with a version of Kubernetes that it does not support, you are using it at your own risk.
+
 ## Acknowledgement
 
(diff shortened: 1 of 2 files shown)


docs/Release_Management.md
```diff
@@ -44,6 +44,18 @@ This project strictly follows [semantic versioning](https://semver.org/spec/v2.0
...
+
+Applicable fixes, including security fixes, may be cherry-picked into the release branch, depending on severity and feasibility. Patch releases are cut from that branch as needed.
+
+We expect users to stay reasonably up-to-date with the versions of Secrets Store CSI Driver they use in production, but understand that it may take time to upgrade. We expect users to be running approximately the latest patch release of a given minor release and encourage users to upgrade as soon as possible.
+
+We expect to "support" n (current) and n-1 major.minor releases. "Support" means we expect users to be running that version in production. For example, when v1.3.0 comes out, v1.1.x will no longer be supported for patches and we encourage users to upgrade to a supported version as soon as possible.
+
+## Supported Kubernetes Versions
+
+Secrets Store CSI Driver is assumed to be compatible with n-4 versions of the latest stable Kubernetes release per [Kubernetes Supported Versions policy](https://kubernetes.io/releases/version-skew-policy/). If you choose to use Secrets Store CSI Driver with a version of Kubernetes that it does not support, you are using it at your own risk.
```
tam7t, line 57: I'm not seeing how the version skep policy link translates to n-4? is that the correct link?

docs/book/src/getting-started/installation.md
```diff
@@ -13,7 +13,9 @@ The helm chart repository URL has changed to `https://kubernetes-sigs.github.io/
 
 #### Supported kubernetes versions
 
-Recommended Kubernetes version:
+The minimum supported Kubernetes version of Secrets Store CSI Driver is n-4 versions of the latest stable Kubernetes release per [Kubernetes Supported Versions policy](https://kubernetes.io/releases/version-skew-policy/).
+
+##### Minimum Kubernetes versions
 
 - **v1.16.0+** (For Linux)
```
tam7t, line 20: According to https://kubernetes.io/releases/patch-releases/ it seems like Kubernetes only supports 1.19 and newer currently.

Should we update accordingly? I know it _should_ work on 1.16 but may be better for us be more proactive in walking that forward as we only have e2e tests for 19, 20, 21, and 22.

docs/Release_Management.md
aramase: > The Kubernetes project maintains release branches for the most recent three minor releases (1.22, 1.21, 1.20). Kubernetes 1.19 and newer receive approximately 1 year of patch support. Kubernetes 1.18 and older received approximately 9 months of patch support.

The current release is 1.22 and 1.19 above is still supported.

tam7t: Ah, I took that as that the intent was to support 3 minor releases and that 1.19 was a special case where it transitioned to a 12 month support flow, but that going forward it will only be 3 branches.

It might be better to phrase it as `Secrets Store CSI Driver will maintain support for all actively supported Kubernetes minor releases`, so that it stays in sync if K8s support periods change

docs/book/src/getting-started/installation.md
nilekhc: Rather than updating versions here every time, I think we should remove section entirely or just point to `supported releases` section. WDYT?

tam7t: That works for me too

Issue #636, opened 2021-07-15
Additional Linters


tam7t:
**Describe the solution you'd like**
Propose expanding `golangci-lint` linters to include the following:

- [ ] errcheck
- [ ] staticcheck
- [ ] typecheck

These would bring us to include [all default](https://golangci-lint.run/usage/linters/).

Additionally I think the following are probably worth including to enforce style and catch bugs:

- [ ] stylecheck
- [x] durationcheck
- [ ] errorlint
- [x] exhaustive
- [x] exportloopref
- [ ] nilerr
- [x] makezero

**Anything else you would like to add:**

I've marked out ones where there are currently no issues and the linter could be added immediately.

The current config has implicit `exclude-use-default: true` which filters out a number of existing findings that should maybe be reviewed individually.

**Environment:**

- Secrets Store CSI Driver version: (use the image tag):
- Kubernetes version: (use `kubectl version`):

aramase:
Let's add `goconst` and `gocyclo` too to the list

jayesh-srivastava:
Hi. I am a new contributor and I would like to work on this issue. Can you please guide me through this? And can you please assign me?

KamalSinghKhanna:
hey , I would like to take this issue suggest me ways to proceed.

tam7t:
This issue should stay open as there are a number of additional linters that would be good to enable.

aramase:
The last PR for the changes was merged: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/790.

/close

PR #790, opened 2021-10-18
chore: add more linters and fix issues

mallow111:
Add the last linter check items
**Which issue(s) this PR fixes
Fixes #https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/636

Added //gocyclo:ignore for function reconcile, the reason is that this is the only function that violates  gocyclo check, also this function itself is pretty complicated, and I am new to this code base, dont want to break it for 1 simple linter check.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,8 +11,10 @@ linters:
     - durationcheck
     - exportloopref
     - errcheck
+    - errorlint
     - exhaustive
     - goconst
+    - gocyclo
     - gofmt
     - goimports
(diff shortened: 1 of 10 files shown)


mallow111:
@arahamad  @tam7t  do you guys still want to keep the errlint then?

tam7t:
/lgtm

I think I like this as-is and we can separately consider the pkg/errors usage

aramase:
/retitle chore: add more linters and fix issues

PR #805, opened 2021-11-10
feat: add token requests client

aramase:

- Adds token request client to the driver
  - This client will be used to generate token for configured audiences and expiration seconds. If this is enabled, providers no longer need to generate the token.
  - We're reusing the token generation logic from kubelet which caches the token until it expires and cache cleanup is explicitly called as part of `NodeUnpublishVolume`.
  - This change is added as backward compat to support kubernetes < 1.20. In 1.20+ by having token requests as part of CSI driver spec, kubelet will generate the token and pass to the driver as part of CSI volume mount context.
  - We're using the CSI driver spec to define the token audience and expiration seconds as mentioned here: https://kubernetes-csi.github.io/docs/token-requests.html
- Added e2e test with e2e-provider 

Fixes #


aramase:
/uncc @ritazh
/cc @nilekhc 

This is ready for review!

config/rbac/role.yaml
```diff
@@ -21,6 +21,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  - create
```
tam7t, line 29: this is pretty sensitive and in past we've done other rbac roles for it specifically... we may want to consider that here too (similar to being able to get all Secrets in a cluster)...

go.mod
```diff
@@ -56,24 +60,56 @@ require (
...
+	k8s.io/apiextensions-apiserver v0.23.0 // indirect
+	k8s.io/apiserver v0.23.0 // indirect
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
+	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
+	sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+replace (
```
tam7t, line 87: i _feel_ like there must be a way to do this without all these replace directives (I'm goin got checkout the branch and try playing around with some things to make `go mod` happy...)

pkg/k8s/token.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		tr, err := c.GetServiceAccountToken(namespace, serviceAccountName, tr)
+		if err != nil {
+			return nil, err
+		}
+		outputs[audience] = tr.Status
+	}
+
+	klog.V(4).InfoS("Fetched service account token attrs for CSIDriver", "driver", c.driverName)
```
tam7t, line 130: may be helpful to have pod UID here too

test/e2eprovider/server/server_test.go
```diff
@@ -213,3 +213,33 @@ func TestRotation(t *testing.T) {
 		t.Errorf("didn't get expected results: (-want +got):\n%s", diff)
 	}
 }
+
+func TestValidateTokens(t *testing.T) {
+	tests := []struct {
+		name      string
+		tokens    string
+		audiences string
+		wantErr   bool
```
tam7t, line 222: Suggest just making 1 passing test and a `TestValidateTokensErrors` for the error case (based on https://github.com/golang/go/wiki/TestComments#table-driven-tests-vs-multiple-test-functions)

tam7t (commented):
Looking great, gonna need a bit more time to think a few things through though!

config/rbac/role.yaml
aramase: sounds good! let me make this a separate file with markers and we can have the user decide if they want to deploy. For helm we can make it conditional based on the `tokenRequests` field.

aramase: done!

test/e2eprovider/server/server_test.go
aramase: done!

go.mod
tam7t: Okay so I think i understand this a bit better...

This PR introduces a dependency on https://pkg.go.dev/k8s.io/kubernetes/pkg/kubelet/token

which pulls in the  `k8s.io/kubernetes v1.23.0` module, but according to https://github.com/kubernetes/kubernetes/issues/90358#issuecomment-617859364

"depending on k8s.io/kubernetes directly as a library, which is not supported"

I guess what we're doing here is whats mentioned https://github.com/kubernetes/kubernetes/issues/79384

Since this token dependency is not permanent (we have plan to remove and replace with `RequiresRepublish` once thats supposed by all K8s supported releases) I think I'm good with these `replace` directives - but we should definitely add a comment explaining that `k8s.io/kubernetes` dependency causes the problems linking to https://github.com/kubernetes/kubernetes/issues/79384

pkg/k8s/token.go
```diff
@@ -0,0 +1,157 @@
...
+	csiDriverLister   storagelisters.CSIDriverLister
+	manager           *token.Manager
+}
+
+// NewTokenClient creates a new TokenClient
+// The client will be used to request a token for the preconfigured audiences (--audiences) and expiration time.
+func NewTokenClient(kubeClient kubernetes.Interface, driverName string) *TokenClient {
+	kubeInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(
+		kubeClient,
+		time.Minute*10,
```
tam7t, line 51: making this a const and/or adding a comment about why 10 minutes would help.

i.e. should this be configurable like the `audience`?

test/e2eprovider/go.mod
```diff
@@ -7,21 +7,51 @@ replace sigs.k8s.io/secrets-store-csi-driver => ../..
...
+	golang.org/x/net v0.0.0-20211209124913-491a49abca63 // indirect
+	golang.org/x/sys v0.0.0-20211029165221-6e7872819dc8 // indirect
+	golang.org/x/text v0.3.7 // indirect
+	google.golang.org/genproto v0.0.0-20210831024726-fe130286e0e2 // indirect
+	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace (
```
tam7t, line 29: oof i'm noticing that this doesnt directly depend on `k8s.io/kubernetes` but still needs all these replace directives...

I wonder if providers (which depend on `sigs.k8s.io/secrets-store-csi-driver` to pull in the protobuf interface) will also need the replace directives. If so we may wanna move the protobuf to its own go module so it doesnt pull in all the transitive dependency issues.

tam7t:
Looking good! just a few more comments

pkg/k8s/token.go
aramase: This is the resync period for the informer. The resync plays back all the events held in the informer cache. This is a controller setting for informer cache and is not something we want to expose to the user.  I can make this configurable by the caller similar to https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/c79c919abae3df5ad7c542f67082ec46e05b538b/pkg/k8s/store.go#L57

test/e2eprovider/go.mod
aramase: opened an issue for it: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/852

PR #810, opened 2021-11-12
Issue #809 - Properly parameterize providersDir values in helm chart

### LINKED ISSUE #809: Helm Chart "linux.providersDir" is not properly used everywhere

Use helm chart

It to support overriding the linux.providersDir value

AWS

**Environment:**

You can see hardcoded values here:
https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L79
https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L140

nccurry:

Fixes #809 

### DIFF
--- a/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
+++ b/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver-windows.yaml
@@ -76,7 +76,7 @@ spec:
             {{- end }}
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
-            - "--provider-volume=C:\\k\\secrets-store-csi-providers"
+            - "--provider-volume={{ .Values.windows.providersDir }}"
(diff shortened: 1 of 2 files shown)


nccurry:
I signed it

aramase (changes requested):
@nccurry Thank you for the PR. The volume mount path is the path where the socket will be available inside the container. The path inside the container set to `/etc/kubernetes/secrets-store-csi-driver-providers` doesn't have to be the same as the actual volume path on the host. Curious what's the motivation for this change. Are you seeing any issues/ this is for consistency?

Also, all the chart changes need to be in the [manifest_staging/](https://github.com/kubernetes-sigs/secrets-store-csi-driver/tree/main/manifest_staging/) folder, where we host the staging charts and deployment yamls. All the yaml changes will then be promoted into the released charts folder with the next release.

nccurry:
@aramase Ah, that's true. I just saw that the path wasn't templated and assumed they needed to match. 

Let me test to verify it works with them being different and I'll close the issue / pr.

PR #824, opened 2021-12-16
Move default CSI providers dir to /var/run

### LINKED ISSUE #823: [Bug] providersDir should not default to `/etc/kubernetes` as primarily contains .sock files better held in `/var/run` or `/tmp/`
When deployed the Secrets Store CSI Driver creates a socket in [/etc/kubernetes/secrets-store-csi-providers](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/manifest_staging/charts/secrets-store-csi-driver/values.yaml#L71), the `/etc/` path is really meant for [configuration files](https://www.linux.com/training-tutorials/linux-directory-structure-explainedetc-folder/), not unix sockets.  The more standard place to put socket files is `/var/run/` or somewhere in `/tmp`  If you explore a debian, redhat, or ubuntu box you'll find most sockets are in one of the two.

This can be a problem for Operating System and Security Controls that are protective of a systems /etc contents.

While this can be overridden in the helm chart by setting `linux.providersDir` the default is re-used in various providers, not all have a method to override it without manually editing the resources.

I would recommend modifying the default and cutting a major release to signify a breaking change to downstream projects.

paulczar:
Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

/var/run is a more canonical place to create unix socket files.  Some
Operating Systems (Red Hat Container Linux, CoreOS, etc) are unfriendly
towards surprise files in /etc, as are many enterprise security
controls.

Fixes #823


tam7t:
From our community call we discussed this change and ways to do it in a backwards compatible way. I've started a PR at https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/851 for that that I'm hoping to finish here shortly(remaining issue is the templating on helm for the additional parameters).

aramase:
Thank you for the PR! 

We have merged https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/851 to make this transition in a backwards compatible way (ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/824#issuecomment-1034004612).

I'm going to close this PR in favor of #851.

/close

PR #836, opened 2022-01-11
helm: additional cluster roles for secretproviderclasses

anapsix:
**What this PR does / why we need it**:
At the moment, there are no ClusterRoles created for SecretProviderClasses resources with aggregation to view and admin default roles. This prevents ClusterRole/admin and ClusterRole/view from accessing SecretProviderClasses resources.

**Which issue(s) this PR fixes**:
Fixes #835

PR #845, opened 2022-01-28
test: add test for view and admin cluster role

anapsix:
As a follow-up to  #836, adding simple tests to verify the new
aggregation roles are applied


aramase (commented):
Thanks for the PR! The cluster roles don't need to be duplicated for e2e but rather we want to validate the file in `manifest_staging`. You can add the deploy commands here: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/d32ca72038650c79561092dab26bf6d5a9c9e40a/Makefile#L403-L408

aramase:
/test: add test for view and admin cluster role

aramase:
/retitle test: add test for view and admin cluster role

anapsix:
Thanks @aramase, done!

aramase (commented):
/lgtm

anapsix:
Should I the same tests to the `test/bats/vault.bats`?

aramase:
> Should I the same tests to the test/bats/vault.bats?

No, just `e2e-provider` is sufficient. Eventually we'll be trimming the provider tests to include basic provider validation only.

PR #873, opened 2022-02-17
fix: resolve grpc error `proto: Marshal called with nil` due to err shadowing

aramase:

As part of the linting changes, we introduced an error check in the defer function: https://github.com/kubernetes-sigs/secrets-store-csi-driver/commit/ca764ea82e37ca6b64c4e559135d453139d5156b. 

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/3eb357e14ec326c80e7a3bcd9aa33b54a47e282f/pkg/secrets-store/nodeserver.go#L82-L84

This defer func and if loop is only triggered when the mount is done and `NodePublishVolume` fails because of provider errors. The actual error in defer func is shadowed by the error returned from `Unmount` which is nil. As a result, the `NodePublishVolume` returns `nil, nil` in case of an actual error which is incorrect. 

- Added a unit test to validate the current behavior is erroneous
- Added the fix in a new commit to validate it fixes the bug

/kind bug

Fixes #

### DIFF
--- a/pkg/secrets-store/nodeserver_test.go
+++ b/pkg/secrets-store/nodeserver_test.go
@@ -322,6 +322,56 @@ func TestNodePublishVolume(t *testing.T) {
 	}
 }
 
+func TestTestNodePublishVolume_ProviderError(t *testing.T) {
+	s := scheme.Scheme
+	s.AddKnownTypes(schema.GroupVersion{Group: secretsstorev1.GroupVersion.Group, Version: secretsstorev1.GroupVersion.Version},
+		&secretsstorev1.SecretProviderClass{},
+		&secretsstorev1.SecretProviderClassList{},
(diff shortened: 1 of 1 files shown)


aramase:
- Added a unit test to validate the current behavior is erroneous https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/873/commits/c452ac46593138a4c4a56fd8fa2ad8e1843b3d33
  - Failing unit test: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_secrets-store-csi-driver/873/pull-secrets-store-csi-driver-unit/1494215488211783680
- Added the fix in a new commit to validate it fixes the bug https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/873/commits/830d184a407c0f7fb9432fcf761de04b292d79b5

PR #957, opened 2022-05-24
docs: include security explanations for root/privileged/and pod tokens

### LINKED ISSUE #707: Document the effects of enabling swap in 1.22+
The driver writes secrets to `tmpfs` volumes [[0]] so that they are kept in memory and not written to disk. That eliminates an important threat vector because the secrets are typically written to volumes unencrypted, and having those on persistent storage is much less desirable than in volatile memory.

In Linux-world, `tmpfs` mounts reside primarily in memory, but can also be swapped onto disk [[1]]. However, for 1.21 and earlier, kubelet would refuse to start up if swap is enabled on a node, so all `tmpfs` volumes were guaranteed to stay only in memory. There's a new alpha feature in 1.22 that allows swap space to be enabled per node [[2]], and so I'm opening this issue because it might be nice to give some guidance on how to manage any security risk that could introduce when used in conjunction with the secrets store CSI driver.

[0]: https://secrets-store-csi-driver.sigs.k8s.io/getting-started/usage.html#secret-content-is-mounted-on-pod-start
[1]: https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html
[2]: https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/

I'm not sure how Windows handles this. Any extra context others can add about Windows would be useful.

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): All versions
- Kubernetes version: (use `kubectl version`): 1.22+

### LINKED ISSUE #763: Document the difference in linux and windows volumes mounts
In case of Linux, the volume is mounted as `tmpfs`. However for windows this is just written to the filesystem as the mount `tmpfs` for windows just creates the directory (target path) if it already doesn't exist. It'll be good to update it here: https://secrets-store-csi-driver.sigs.k8s.io/concepts.html#how-it-works

/kind documentation

tam7t:
Also addresses issues around accidental persistence of secret material
to disks #707 (document swap) and #763 (windows).

/kind documentation

Fixes #707, Fixes #763

### DIFF
--- a/docs/book/src/concepts.md
+++ b/docs/book/src/concepts.md
@@ -31,6 +31,27 @@ Currently supported providers:
 - [GCP Provider](https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp)
 - [Vault Provider](https://github.com/hashicorp/secrets-store-csi-driver-provider-vault)
 
+## Security
+
+The Secrets Store CSI Driver **daemonset** runs as `root` in a `privileged` pod. This is because the **daemonset** is
+responsible for creating new `tmpfs` filesystems and `mount`ing them into existing pod filesystems within the node's
+`hostPath`. `root` is necessary for the `mount` syscall and other filesystem operations and `privledged` is required for
+to use `mountPropagation: Bidirectional` to modify other running pod's filesystems.
+
+The provider plugins are also required to run as `root` (though `privileged` should not be necessary). This is because
+the provider plugin must create a unix domain socket in a `hostPath` for the driver to connect to.
+
+Further, service account tokens for pods that require secrets may be forwarded from the kubelet process to the driver
+and then to provider plugins. This allows the provider to impersonate the pod when contacting the external secret API.
+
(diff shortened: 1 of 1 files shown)


docs/book/src/concepts.md
```diff
@@ -31,6 +31,27 @@ Currently supported providers:
...
 
+## Security
+
+The Secrets Store CSI Driver **daemonset** runs as `root` in a `privileged` pod. This is because the **daemonset** is
+responsible for creating new `tmpfs` filesystems and `mount`ing them into existing pod filesystems within the node's
+`hostPath`. `root` is necessary for the `mount` syscall and other filesystem operations and `privledged` is required for
+to use `mountPropagation: Bidirectional` to modify other running pod's filesystems.
+
+The provider plugins are also required to run as `root` (though `privileged` should not be necessary). This is because
+they provider plugin must create a unix domain socket in a `hostPath` for the driver to connect to.
```
aramase, line 42:
```suggestion
the provider plugin must create a unix domain socket in a `hostPath` for the driver to connect to.
```

PR #1013, opened 2022-07-27
fix: sanitize service account tokens in logs

aramase:

**What type of PR is this?**

/kind bug

`protosanitizer` strips out sensitive information from the secrets field in the RPC request, however it doesn't handle
service account tokens in the volume context. Added a new wrapper function to sanitize sensitive information before sending it to the protosanitizer.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/rotation/reconciler.go
+++ b/pkg/rotation/reconciler.go
@@ -55,18 +55,12 @@ import (
 )
 
 const (
-	permission       os.FileMode = 0644
-	maxNumOfRequeues int         = 5
+	maxNumOfRequeues int = 5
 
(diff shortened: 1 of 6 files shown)


nilekhc:
@aramase Would it be helpful to add this directly upstream?

/lgtm

aramase:
> @aramase Would it be helpful to add this directly upstream?

I looked at it but that's not possible. `protosanitizer` works based on field tags and `VolumeContext` field is a map. The only way we can do it upstream is by marking `VolumeContext` as secret and every value in that will be stripped in logs (including pod name, namespace, service account name).

PR #1047, opened 2022-09-09
Fixed hook truncation, moved name overrides to globals

### LINKED ISSUE #1046: Move fullNameOverride to a global value and address truncation issues
Currently, `.Values.fullnameOverride` is used by this chart when generating many of the names in the helper templates. However, downstream charts lik the [azure csi secrets driver](https://github.com/Azure/secrets-store-csi-driver-provider-azure) also use the same value inside of some of its helpers. Therefore, in order to get the expected functionality when using the subchart, one has to override its field using `.Values.fullnameOverride` as well as this chart's field using `.Values.secrets-store-csi-driver.fullnameOverride`. To me, this is a perfect candidate for using [helm's global values section(https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values). The same field can be referenced by both charts using the `.Values.global.fullnameOverride`. Conceptually, it makes sense that this value should only have to be set once. 

I am submitting a corresponding issue request for the azure driver downstream chart which I will link here. Thanks

Additionally, in terms of truncation issues, the crd hooks don't take into account the need to truncate their added suffix. See this example from the CRDs upgrade hook:
```
{{ template "sscd.fullname" . }}-upgrade-crds
```
There should be additional truncation of the sscd.fullname of 63 minus the length of -upgrade-crds. I will add this logic in as well.
- Secrets Store CSI Driver version: (use the image tag): N/A
- Kubernetes version: (use `kubectl version`): N/A

miles-w-3:
**What type of PR is this?**

/kind bug
- Ensures truncation is properly set at 63, for crd hooks, previously unaddressed because suffixes were added
- Adds `global nameOverride`,` fullnameOverride`, and `commonLabels` variables so downstream charts like azure can set one variable instead of setting it twice
  - Still supports setting old variable for backwards compatibility
Fixes #1046 

This is a chart update. It should not affect the yaml except for truncation cases that would have previously failed on apply anyway. (When `fullnameOverride` was already 63 chars long and got the suffix appended onto it)

**TODOs**:

- [ X ] squashed commits
- [ X ] includes documentation
- [  ] adds unit tests


aramase (commented):
Thanks for the PR @miles-w-3. Could you sing [CLA](https://github.com/kubernetes/community/blob/master/CLA.md)?

aramase:
@miles-w-3 Please see https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1047#pullrequestreview-1104875448 for CLA.

I'm closing this PR now due to inactivity. Please feel free to reopen if it's still being worked on. 

/close

PR #1058, opened 2022-10-06
fix: handles pfx certs in k8s secrets sync

### LINKED ISSUE #1057: SPC with synced K8S Secret populating the secret with an "empty" cert
We have a fairly standard app consisting of a Deployment, Service, SPC w/ synced Secrets, and Ingress. The SPC is mounted on the deployment pods and is syncing secrets that the ingress references as TLS certs. However, the synced secrets' data sections have only empty values:
```yaml
data:
  tls.crt: ""
  tls.key: <base64-encoded "---BEGIN PRIVATE KEY---\n---END PRIVATE KEY---". Just those lines, no content.>
```

The referenced certificate to be synced successfully

So as described above, we have an app with a fairly standard setup of ingresses, and when deploying it, we're hitting a rather strange scenario where it's syncing an "empty" secret. Specs included below, as they're rather large - in essence, an SPC has a _somewhat_ large (about 8) set of synced objects it's using for creating a set of ingress secrets. the SPC is mounted on the pod and works fine - the pod starts up successfully - but the secrets that are synced are invalid.

In particular, the equivalent of ` notebook-instance-proxy-managed-ingress-ssl-synced` in our specs is empty, with the contents I pasted above:
```yaml
data:
  tls.crt: ""
  tls.key: <base64-encoded "---BEGIN PRIVATE KEY---\n---END PRIVATE KEY---">
```

I've checked through a wide variety of logs from both csi and azure components - there were no error logs, and all logs mentioning the pod name or SPC indicated nothing went wrong or differently than any of our other deployments that use CSI Driver.

It doesn't appear to be like a one off transient issue, as deployments of that same build to a different namespace (for PR gates) also have the same issue. deleting and redeploying the app to recreate the secrets has resulted in the same issue as well. We're kind of at a loss for what might be going on.

The only thing that's stood out to me so far is that the SPC in question has a relatively large number of separately-synced objects.

One other notable fact might be that we've migrated from self-hosted to AKS-hosted CSI driver, in case there might be some strangeness with that handover.

Spec note - these approximate the shape of what we're doing with some stuff omitted for brevity or security. I can privately send over a more complete set of files if needed.
```yaml
# Deployment
apiVersion: apps/v1
kind: Deployment
metadata:
  # [... Truncated ...]
spec:
  progressDeadlineSeconds: 600
  replicas: 10
  revisionHistoryLimit: 3
  selector:
    matchLabels:
      app: notebook-instance-proxy
  strategy:
    rollingUpdate:
      maxSurge: 10%
      maxUnavailable: 10%
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: notebook-instance-proxy
        app-web: notebook-instance-proxy
      namespace: vienna-master
    spec:
      containers:
        # [ ... Truncated ... ]
        volumeMounts:
        - mountPath: /app/secrets/
          name: secrets-volume
          readOnly: true
      # [ ... Truncated ... ]
      volumes:
      - csi:
          driver: secrets-store.csi.k8s.io
          nodePublishSecretRef:
            name: notebook-instance-proxy-kv-creds
          readOnly: true
          volumeAttributes:
            secretProviderClass: notebook-instance-proxy-secretprovider
        name: secrets-volume

---
# SPC
apiVersion: secrets-store.csi.x-k8s.io/v1
kind: SecretProviderClass
metadata:
  name: notebook-instance-proxy-secretprovider
  namespace: vienna-master
spec:
  parameters:
    cloudEnvFileName: /
    cloudName: AZUREPUBLICCLOUD
    keyvaultName: <ommitted>
    objects: |
      array:
        - |
          objectName: "cert-1"
          objectAlias: "cert-1"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-2"
          objectAlias: "cert-2"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-3"
          objectAlias: "cert-3"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-4"
          objectAlias: "cert-4"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-5"
          objectAlias: "cert-5"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-6"
          objectAlias: "cert-6"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-7"
          objectAlias: "cert-7"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "cert-8"
          objectAlias: "cert-8"
          objectType: secret
          objectFormat: pfx
          objectEncoding: base64
        - |
          objectName: "secret-1"
          objectAlias: "secret-1"
          objectType: secret
        - |
          objectName: "secret-2"
          objectAlias: "secret-2"
          objectType: secret
        - |
          objectName: "secret-3"
          objectAlias: "secret-3"
          objectType: secret
        - |
          objectName: "secret-4"
          objectAlias: "secret-4"
          objectType: secret
        - |
          objectName: "secret-5"
          objectAlias: "secret-5"
          objectType: secret
        - |
          objectName: "secret-6"
          objectAlias: "secret-6"
          objectType: secret
    tenantId: <ommitted>
  provider: azure
  secretObjects:
  - data:
    - key: tls.crt
      objectName: cert-1
    - key: tls.key
      objectName: cert-1
    secretName: notebook-instance-proxy-managed-ingress-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: tls.crt
      objectName: cert-2
    - key: tls.key
      objectName: cert-2
    secretName: notebook-instance-proxy-ingress-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: tls.crt
      objectName: cert-3
    - key: tls.key
      objectName: cert-3
    secretName: notebook-instance-proxy-modern-ingress-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: tls.crt
      objectName: cert-4
    - key: tls.key
      objectName: cert-4
    secretName: notebook-instance-proxy-managed-ms-ingress-internal-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: tls.crt
      objectName: cert-5
    - key: tls.key
      objectName: cert-5
    secretName: notebook-instance-proxy-nodes-ingress-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: tls.crt
      objectName: cert-6
    - key: tls.key
      objectName: cert-6
    secretName: notebook-instance-proxy-certapi-ingress-ssl-synced
    type: kubernetes.io/tls
  - data:
    - key: secret-1
      objectName: secret-1
    - key: secret-2
      objectName: secret-2
    - key: secret-3
      objectName: secret-3
    - key: secret-4
      objectName: secret-4
    - key: secret-5
      objectName: secret-5
    - key: secret-6
      objectName: secret-6
    secretName: notebook-instance-proxy-synced
    type: Opaque

---

# Ingress
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    # [... ommitted ...]
  generation: 48
  name: notebook-instance-proxy-ingress
  namespace: vienna-master
spec:
  rules:
  - host: auth.master.instances.azureml-test.ms
    http:
      paths:
      - backend:
          service:
            name: notebook-instance-proxy
            port:
              number: 443
        path: /
        pathType: Prefix
  - host: '*.master.instances.azureml-test.net'
    http:
      paths:
      - backend:
          service:
            name: notebook-instance-proxy
            port:
              number: 443
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - '*.master.instances.azureml-test.net'
    secretName: notebook-instance-proxy-managed-ingress-ssl-synced
  - hosts:
    - auth.master.instances.azureml-test.ms
    - '*.master.instances.azureml-test.ms'
    secretName: notebook-instance-proxy-modern-ingress-ssl-synced
```

Azure Key Vault

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): AKS-managed
- Kubernetes version: (use `kubectl version`): 1.23

nilekhc:
Signed-off-by: Nilekh Chaudhari <1626598+nilekhc@users.noreply.github.com>

**What type of PR is this?**

/kind bug

Fixes #1057

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/util/secretutil/secret.go
+++ b/pkg/util/secretutil/secret.go
@@ -30,6 +30,7 @@ import (
 
 	secretsstorev1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
 
+	"golang.org/x/crypto/pkcs12"
 	corev1 "k8s.io/api/core/v1"
 )
 
@@ -40,7 +41,7 @@ const (
 	privateKeyTypeEC  = "EC PRIVATE KEY"
 )
 
-// getCertPart returns the certificate or the private key part of the cert
+// GetCertPart returns the certificate or the private key part of the cert
 func GetCertPart(data []byte, key string) ([]byte, error) {
 	if key == corev1.TLSPrivateKeyKey {
 		return getPrivateKey(data)
@@ -65,16 +66,34 @@ func getCert(data []byte) ([]byte, error) {
 		}
 		data = rest
 	}
+
+	// if cert is nil, then it might be a pfx cert
+	if certs == nil {
+		pemBlocks, err := pkcs12.ToPEM(data, "")
+		if err != nil {
+			return nil, err
+		}
+
+		// pem Blocks returns both the certificate and private key types
+		for _, block := range pemBlocks {
+			// get bytes for certificate
+			if block.Type == certType {
+				certs = append(certs, pem.EncodeToMemory(block)...)
+			}
+		}
+	}
+
 	return certs, nil
 }
 
 // getPrivateKey returns the private key part of a cert
 func getPrivateKey(data []byte) ([]byte, error) {
-	var der, derKey []byte
+	var der, derKey, rest []byte
+	var pemBlock *pem.Block
 	privKeyType := privateKeyType
 
 	for {
-		pemBlock, rest := pem.Decode(data)
+		pemBlock, rest = pem.Decode(data)
 		if pemBlock == nil {
 			break
 		}
@@ -82,6 +101,22 @@ func getPrivateKey(data []byte) ([]byte, error) {
 			der = pemBlock.Bytes
 		}
 		data = rest
+	}
+
+	// if both der is nil, then certificate might be in the pfx format
+	if der == nil {
+		pemBlocks, err := pkcs12.ToPEM(data, "")
+		if err != nil {
+			return nil, err
+		}
+
+		// pem blocks returns both the certificate and private key types
+		for _, block := range pemBlocks {
+			// get bytes for private key
+			if block.Type == privateKeyType {
+				der = block.Bytes
+			}
+		}
 	}
 
 	// parses an RSA private key in PKCS #1, ASN.1 DER form
--- a/pkg/util/secretutil/secret_test.go
+++ b/pkg/util/secretutil/secret_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package secretutil
 
 import (
+	"encoding/base64"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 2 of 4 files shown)


jroggeman (approved):

pkg/util/secretutil/secret.go
```diff
@@ -28,6 +28,7 @@ import (
 	"sort"
 	"strings"
 
+	"golang.org/x/crypto/pkcs12"
```
aramase, line 31: Move this after line 34 and group with `corev1 "k8s.io/api/core/v1"` as they're external packages

pkg/util/secretutil/secret.go
```diff
@@ -84,6 +102,19 @@ func getPrivateKey(data []byte) ([]byte, error) {
 		data = rest
 	}
 
+	// if both der and pemBlock are nil, then it might be a pfx cert
```
aramase, line 105:
```suggestion
	// if both der and pemBlock are nil, then it might be in the pfx format
```

pkg/util/secretutil/secret.go
```diff
@@ -84,6 +102,19 @@ func getPrivateKey(data []byte) ([]byte, error) {
...
+	if der == nil && pemBlock == nil {
```
aramase, line 106: The `for` loop only breaks when `pemBlock == nil` is true. Is the `pemBlock == nil` check here redundant? If yes, can we remove it?

pkg/util/secretutil/secret.go
```diff
@@ -84,6 +102,19 @@ func getPrivateKey(data []byte) ([]byte, error) {
...
+		key, _, err := pkcs12.Decode(data, "")
```
aramase, line 107: If we use `der` instead of new variable key, couldn't we rely on existing key parsing logic to marshal based on the key type (same as today)?

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/b2631980e2790b102dfb78afa4c90fa9c3c83fff/pkg/util/secretutil/secret.go#L87-L115

pkg/util/secretutil/secret.go
```diff
@@ -84,6 +102,19 @@ func getPrivateKey(data []byte) ([]byte, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode pkcs12 data, err: %w", err)
+		}
+
+		derKey, err = x509.MarshalPKCS8PrivateKey(key.(*rsa.PrivateKey))
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal private key, err: %w", err)
+		}
```
aramase, line 115: same comment as above (instead of `MarshalPKCS8PrivateKey`, could we reuse the existing key parse logic?)

pkg/util/secretutil/secret_test.go
```diff
@@ -168,6 +169,53 @@ func TestGetCert(t *testing.T) {
 	}
 }
 
+func TestPfxGetCert(t *testing.T) {
```
aramase, line 172: Instead of a new test, I would recommend adding a new table test in [`TestGetCert`](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/pkg/util/secretutil/secret_test.go#L133)

pkg/util/secretutil/secret_test.go
```diff
@@ -597,6 +645,63 @@ g+Ia2YI15BzapW0agqSSTlfGMoQHaPRh1+XYtkOd/xb4xc8d+gc0
 	}
 }
 
+func TestPfxGetPrivateKey(t *testing.T) {
```
aramase, line 648: Instead of a new test, I would recommend adding a new table test in [`TestGetPrivateKey `](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/pkg/util/secretutil/secret_test.go#L398)

aramase (changes requested):

pkg/util/secretutil/secret.go
nilekhc: With `pkcs12.Decode` we directly get the private key and don't have to parse it. So I am directly marshaling it and assigning it to `derKey`. Let me know if I am missing something here.

pkg/util/secretutil/secret.go
nilekhc: Same as above. IMO we don't need to parse it.

pkg/util/secretutil/secret_test.go
nilekhc: I thought about it. But for PFX I am having an input in base64 encoded format and need to decode it before we pass it to the function. It's not needed for existing tests and hence created a new test to keep it clean.

pkg/util/secretutil/secret_test.go
nilekhc: Same as other test explanations.

pkg/util/secretutil/secret.go
aramase: We get the private key with `pem.Decode` too. The rest of the code here is to parse and determine if there is an error, if not we marshal it with the appropriate type (RSA/EC). By always using `MarshalPKCS1PrivateKey` here, we're assuming it's in PKCS #1 and not PKCS #8, which isn't valid. The parse code handles those scenarios and as part of the PEM block returned, it set the correct key type (`EC` or `RSA`

pkg/util/secretutil/secret_test.go
aramase: You can make actual a function field `func() []byte` and use that as input for `getPrivateKey`. We don't have to worry about the base64 decode error because we pass valid inputs but if we want we can still assert that in the function. Only reason for having all these tests in one is to ensure we always assert the same things for a function unit test.

aramase (approved):
/lgtm
/approve

PR #1086, opened 2022-10-26
fix: escape dot in target path regex

aramase:

**What type of PR is this?**
/kind bug

ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/security/code-scanning/1

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
     - main
+  pull_request:
+    branches:
+    - main
   schedule:
     - cron: "0 7 * * 1" # Mondays at 7:00 AM
 
--- a/pkg/util/fileutil/filesystem_test.go
+++ b/pkg/util/fileutil/filesystem_test.go
@@ -267,6 +267,10 @@ func TestGetVolumeNameFromTargetPath(t *testing.T) {
 			want:       "",
 		},
 		{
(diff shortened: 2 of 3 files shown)


.github/workflows/codeql.yaml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
     - main
+  pull_request:
+    branches:
+    - main
```
aramase, line 9: It only takes ~ 8m to complete (ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/actions/runs/3331865967), so enabling it as part of pull requests

pkg/util/fileutil/filesystem_test.go
```diff
@@ -266,6 +266,10 @@ func TestGetVolumeNameFromTargetPath(t *testing.T) {
 			targetPath: "/var/lib/kubelet/pods/7e7686a1-56c4-4c67-a6fd-4656ac484f0a/volumes/kubernetes.io~pv/pvvol/mount",
 			want:       "",
 		},
+		{
+			targetPath: "/var/lib/kubelet/pods/7e7686a1-56c4-4c67-a6fd-4656ac484f0a/volumes/kubernetesxio~csi/secrets-store-inline/mount",
+			want:       "",
+		},
```
aramase, line 272: This gives `secrets-store-inline` without escaping the `.` which is fixed with this PR. Adding this test to validate the regex does a strict match.

PR #1124, opened 2022-12-18
feat: require RequiresRepublish for rotation

### LINKED ISSUE #585: Using `RequiresRepublish` for rotation
When `requiresRepublish: true` is set in the `CSIDriver` spec

```
  requiresRepublish: true # added in Kubernetes 1.20, this field is beta as of Kubernetes 1.21
```
kubelet will periodically send `NodePublishVolume` to the driver.

- Investigate if we can use this feature with internal backoff logic for rotation

/assign

tam7t:
The flags will be no-ops in this and removed in 1.5+.

/kind deprecation

Remove the rotation controller and rely exclusively on RequiresRepublish for secret rotation.

All supported k8s versions have RequiresRepublish support enabled by default.

Fixes #585 

For release we would need to remember to remove the files and update the release notes to instruct that existing rbac permissions be deleted (I'm not sure whether helm has support for deletion of that)

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


aramase:
Synced with @tam7t on slack. Going to close this PR and i'll open a new one with the changes.

/close

PR #1202, opened 2023-04-02
chore: cleanup WritePayloads to be easier to use

tam7t:
**What type of PR is this?**

/kind cleanup

Makes WritePayloads() call Validate() internally so that it can be used safely on its own. Also cleans up some of the indentation on the provider client.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/secrets-store/provider_client.go
+++ b/pkg/secrets-store/provider_client.go
@@ -66,10 +66,12 @@ const ServiceConfig = `
 `
 
 var (
-	// PluginNameRe is the regular expression used to validate plugin names.
-	PluginNameRe        = regexp.MustCompile(`^[a-zA-Z0-9_-]{0,30}$`)
-	ErrInvalidProvider  = errors.New("invalid provider")
-	ErrProviderNotFound = errors.New("provider not found")
+	// pluginNameRe is the regular expression used to validate plugin names.
+	pluginNameRe = regexp.MustCompile(`^[a-zA-Z0-9_-]{0,30}$`)
+
+	errInvalidProvider       = errors.New("invalid provider")
+	errProviderNotFound      = errors.New("provider not found")
+	errMissingObjectVersions = errors.New("missing object versions")
 )
 
 // PluginClientBuilder builds and stores grpc clients for communicating with
@@ -123,8 +125,8 @@ func (p *PluginClientBuilder) Get(ctx context.Context, provider string) (v1alpha
 	}
 
 	// client does not exist, create a new one
-	if !PluginNameRe.MatchString(provider) {
-		return nil, fmt.Errorf("%w: provider %q", ErrInvalidProvider, provider)
+	if !pluginNameRe.MatchString(provider) {
+		return nil, fmt.Errorf("%w: provider %q", errInvalidProvider, provider)
 	}
 
 	// check all paths
@@ -147,7 +149,7 @@ func (p *PluginClientBuilder) Get(ctx context.Context, provider string) (v1alpha
 	}
 
 	if socketPath == "" {
-		return nil, fmt.Errorf("%w: provider %q", ErrProviderNotFound, provider)
+		return nil, fmt.Errorf("%w: provider %q", errProviderNotFound, provider)
 	}
 
 	conn, err := grpc.Dial(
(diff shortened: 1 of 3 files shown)


pkg/secrets-store/provider_client.go
```diff
@@ -263,18 +263,16 @@ func MountContent(ctx context.Context, client v1alpha1.CSIDriverProviderClient,
...
-			return nil, internalerrors.FileWriteError, err
-		}
-		if err := fileutil.WritePayloads(targetPath, resp.GetFiles()); err != nil {
-			return nil, internalerrors.FileWriteError, err
-		}
-	} else {
+	if len(resp.GetFiles()) == 0 {
 		// when no files are returned we assume that the plugin has not migrated
 		// grpc responses for writing files yet.
 		klog.V(5).Info("mount response has no files")
```
nilekhc, line 269: Could we be more descriptive here? For end users who are seeing this info msg wouldn't make much sense unless they know the history of provider vs driver writing.
IMO it would be good to put a clear message stating what "no files" means.

pkg/secrets-store/provider_client.go
```diff
@@ -263,18 +263,16 @@ func MountContent(ctx context.Context, client v1alpha1.CSIDriverProviderClient,
...
+	} else {
+		klog.V(5).Info("writing mount response files")
```
nilekhc, line 271: Instead of having an else statement just for Info, would it be helpful to move it below after `fileutil.WritePayloads` completes successfully without error?

pkg/secrets-store/provider_client.go
tam7t: Updated the message and added a comment referencing the issue

pkg/secrets-store/provider_client.go
tam7t: Removed the else and just have it always attempt to write the files. If there are no files it should just be a no-op.

pkg/secrets-store/provider_client.go
```diff
@@ -67,9 +67,11 @@ const ServiceConfig = `
...
 var (
 	// PluginNameRe is the regular expression used to validate plugin names.
-	PluginNameRe        = regexp.MustCompile(`^[a-zA-Z0-9_-]{0,30}$`)
-	ErrInvalidProvider  = errors.New("invalid provider")
-	ErrProviderNotFound = errors.New("provider not found")
+	PluginNameRe = regexp.MustCompile(`^[a-zA-Z0-9_-]{0,30}$`)
+
+	ErrInvalidProvider       = errors.New("invalid provider")
+	ErrProviderNotFound      = errors.New("provider not found")
+	ErrMissingObjectVersions = errors.New("missing object versions")
```
aramase, line 74: These error variables are only used in the `secretsstore` package. Can we unexport these?

pkg/secrets-store/provider_client.go
```diff
@@ -263,19 +265,16 @@ func MountContent(ctx context.Context, client v1alpha1.CSIDriverProviderClient,
...
-		}
-	} else {
-		// when no files are returned we assume that the plugin has not migrated
-		// grpc responses for writing files yet.
-		klog.V(5).Info("mount response has no files")
+	if len(resp.GetFiles()) == 0 {
+		// The plugin mount response contains no files. Possible that the plugin
+		// is writing its own files instead of the driver (See Issue #551).
+		klog.V(5).Info("Empty files in mount response. It is possible that the plugin has not migrated to driver-written files (Issue #551).")
+	}
```
aramase, line 272: Could we just return early here?

pkg/secrets-store/provider_client.go
```diff
@@ -263,19 +265,16 @@ func MountContent(ctx context.Context, client v1alpha1.CSIDriverProviderClient,
...
+
+	if err := fileutil.WritePayloads(targetPath, resp.GetFiles()); err != nil {
+		return nil, internalerrors.FileWriteError, err
 	}
+	klog.V(5).Info("mount response files written.")
```
aramase, line 277: Added a comment on returning early above. If we don't return early, these message will conflict with [log](https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1202/files#diff-39f63ba0bd3c5582f2919a5632b3b7d89d55a105ce5aecee4325fc146e8c5205R271) here causing confusion.

pkg/secrets-store/provider_client.go
tam7t: :+1:

pkg/secrets-store/provider_client.go
tam7t: :+1:

pkg/secrets-store/provider_client.go
tam7t: :+1:

aramase (commented):
/lgtm

PR #1204, opened 2023-04-04
docs: update reference to `registry.k8s.io` in release

aramase:
/kind documentation

With the registry change from `k8s.gcr.io` to `registry.k8s.io`, we'll need to update the file in `registry.k8s.io/*` to promote the driver images. This PR updates the release doc with the new script path and updates `k8s.gcr.io` reference to `registry.k8s.io`.


aramase:
> There are a few more references to things like gcr.io/k8s-staging-test-infra, should any of those references be updated or is it really just the release images that changed?

Those can stay the same. The image when promoted will be pushed to `registry.k8s.io` instead of `k8s.gcr.io`.

PR #1254, opened 2023-05-09
chore: bump k8s deps to `v1.26.4`

aramase:
/kind feature

- bump k8s deps to `v1.26.4`
- bump otel to `v0.38.1` and update metrics
- drop the `total_` prefix from the metric name
  - For prometheus counters, by default they're appending _total if it's
monotonic counter. Removing the total_ prefix as it's redundant in the
name (xref: https://github.com/open-telemetry/opentelemetry-go/pull/3360).
This is a breaking change.
- collect metrics as part of CI
- validate metrics as part of CI


aramase:
https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1254/commits/121699b1835259ac8c99e84f09a30d55f6fec784 is a breaking change. Added more description in the commit.

Issue #1281, opened 2023-06-15
Aggregated cluster roles for view access to SecretProviderClassPodStatus


erikgb:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Use case: As a regular namespace user I would like to track the binding between a pod and `SecretProviderClass` (`SecretProviderClassPodStatus`). Access to the status resource currently requires cluster-admin role or custom RBAC. I think secrets-store-csi-driver should by default allow **view** access to aggregates cluster roles: view, edit, admin and cluster-reader.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Aggregated cluster roles for `SecretProviderClass` were added in https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/836, but I don't understand why this wasn't fixed for `SecretProviderClassPodStatus` at the same time....

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): N/A
- Kubernetes version: (use `kubectl version`): N/A

erikgb:
This is what I think should be provisioned by default:

````yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    rbac.authorization.k8s.io/aggregate-to-view: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
  name: secretproviderclasspodstatuses-viewer
rules:
  - apiGroups:
      - secrets-store.csi.x-k8s.io
    resources:
      - secretproviderclasspodstatuses
    verbs:
      - get
      - list
      - watch
````

PR #1283, opened 2023-06-16
helm: aggregated clusterrole to view secretproviderclasspodstatuses

erikgb:
**What type of PR is this?**

This adds a new aggregated cluster role granting access to view `SecretProviderClassPodStatus`.

/kind feature

As a regular user using `SecretProviderClass`, there is currently no RBAC that allows me to view the status of SPC usage in pods. We have worked around this by asking our cluster-admin to create the aggregated clusterrole included in this PR. I think it makes sense to include this cluster role by default in an installation of Secrets Store CSI Driver.

Fixes https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/1281

~I have included aggregation to the `cluster-reader` cluster role. This cluster role is not listed among the[ default user-facing roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) in the Kubernetes docs, but I think it is quite common and makes sense. It is [included by default](https://docs.openshift.com/container-platform/4.13/authentication/using-rbac.html#default-roles_using-rbac) in OpenShift, and we use it quite a lot. I have also added this aggregation to the existing `secretproviderclasses-viewer-role` to make it consistent.~ Update: I will file a separate issue/PR for this.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/manifest_staging/charts/secrets-store-csi-driver/templates/role-secretproviderclasspodstatuses-viewer.yaml
+++ b/manifest_staging/charts/secrets-store-csi-driver/templates/role-secretproviderclasspodstatuses-viewer.yaml
@@ -1,0 +1,21 @@
+{{ if .Values.rbac.install }}
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  labels:
+{{ include "sscd.labels" . | indent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: secretproviderclasspodstatuses-viewer-role
+rules:
+- apiGroups:
+  - secrets-store.csi.x-k8s.io
+  resources:
+  - secretproviderclasspodstatuses
+  verbs:
+  - get
+  - list
+  - watch
+{{ end }}
--- a/manifest_staging/deploy/role-secretproviderclasspodstatuses-viewer.yaml
+++ b/manifest_staging/deploy/role-secretproviderclasspodstatuses-viewer.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: secretproviderclasspodstatuses-viewer-role
+rules:
+- apiGroups:
+  - secrets-store.csi.x-k8s.io
+  resources:
+  - secretproviderclasspodstatuses
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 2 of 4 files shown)


manifest_staging/charts/secrets-store-csi-driver/templates/role-secretproviderclasspodstatuses-viewer.yaml
```diff
@@ -0,0 +1,24 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  labels:
+{{ include "sscd.labels" . | indent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```
aramase, line 12: aggregate to `admin` should be fine but we don't need aggregate to `edit`.

manifest_staging/deploy/role-secretproviderclasspodstatuses-viewer.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
```
aramase, line 6: these should match what's defined in helm chart manifests

aramase (changes requested):

manifest_staging/charts/secrets-store-csi-driver/templates/role-secretproviderclasspodstatuses-viewer.yaml
erikgb: Could you please explain why? IMO users in **either** `view`,  `edit` or `admin` role should be able to view secretproviderclasspodstatuses. Ref. https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles. The aggregation labels should be the same as the aggregation cluster role secretproviderclasses-viewer-role.

erikgb: Ah, the default aggregating user-facing roles are also aggregating each other. Makes sense - I didn't know that, thanks! Fixed now.

aramase (commented):
/lgtm
/approve

PR #1368, opened 2023-11-02
chore: update to go 1.21

aramase:
/kind feature

- update to go 1.21


Makefile
```diff
@@ -161,7 +161,7 @@ image-scan: $(TRIVY)
 	$(TRIVY) image --severity MEDIUM,HIGH,CRITICAL $(CRD_IMAGE_TAG)
 	# show vulnerabilities that have been fixed
 	$(TRIVY) image --exit-code 1 --ignore-unfixed --severity MEDIUM,HIGH,CRITICAL $(IMAGE_TAG)
-	$(TRIVY) image --vuln-type os --exit-code 1 --ignore-unfixed --severity MEDIUM,HIGH,CRITICAL $(CRD_IMAGE_TAG)
```
nilekhc, line 164: Dropping `--vuln-type os` will scan for all types of vulnerabilities?

nilekhc (commented):
Just a quick question. Otherwise LGTM.

Makefile
aramase: that's right! we added `--vuln-type os` because the old `kubectl` binary had CVE and we didn't have a release that contains the fixes.

PR #1469, opened 2024-03-13
Use values.yaml for default values in securityContext

NaMNDV:
**What type of PR is this?**
/kind cleanup

In secret-store-csi-driver.yaml there is a securityContext which uses `privileged: true` and does not use values.yaml to apply the defaults. This PR migrates to the values as default values in values.yaml for an easy configuration
Fixes #
None, there is an open comment in a bug in gcp provider https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/issues/242#issuecomment-1928894797

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
+++ b/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml
@@ -103,7 +103,7 @@ spec:
                 fieldPath: spec.nodeName
           imagePullPolicy: {{ .Values.linux.image.pullPolicy }}
           securityContext:
-            privileged: true
+            privileged: {{ .Values.securityContext.privileged}}
           {{- if semverCompare ">= v0.0.8-0" .Values.linux.image.tag }}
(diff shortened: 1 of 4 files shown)


aramase (changes requested):
`privileged: true` is **required** for the driver to function correctly. Moving this to `values.yaml` could cause the driver functionality to fail if set to false.

I don't see the reason to make a required security context `privileged: true` be a configuration in values.yaml.

This is documented (ref: https://secrets-store-csi-driver.sigs.k8s.io/concepts#security)
> The Secrets Store CSI Driver daemonset runs as root in a privileged pod. This is because the daemonset is responsible for creating new tmpfs filesystems and mounting them into existing pod filesystems within the node’s hostPath. root is necessary for the mount syscall and other filesystem operations and privileged is required for to use mountPropagation: Bidirectional to modify other running pod’s filesystems.

PR #1648, opened 2024-09-27
auto-restart

### LINKED ISSUE #1647: Document auto restart of pods on secret rotation
When secret auto-rotation is enabled, users of Secrets Store CSI Driver may look for a means to have workload pods automatically restarted when a mounted secret is updated. While restart code is assumed not to be in scope of the driver project, the project could outline a suggested approach.

I have implemented a kubernetes controller to optionally restart pods when mounted secrets are updated.

I have a document ready to propose for inclusion in docs/topics.

Note that Reloader does not cover the case where secrets are mounted in pod volumes.

This follow from discussion on the slack community: https://kubernetes.slack.com/archives/C013PUP2WRK/p1727349029443369?thread_ts=1724065955.676549&cid=C013PUP2WRK
**Environment:**

- Secrets Store CSI Driver version: (use the image tag): N/A
- Kubernetes version: (use `kubectl version`): N/A

ewan-chalmers:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**: Provides guidance for users of the auto-rotation feature of Secrets Store CSI Driver on how to implement optional auto-restart of workload pods when secrets are updated.

Fixes #1647

**TODOs**:

- [ ] squashed commits
- [X] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/book/src/topics/auto-restart-on-rotation.md
+++ b/docs/book/src/topics/auto-restart-on-rotation.md
@@ -1,0 +1,120 @@
+# Auto restart of pods when secret is rotated
+
+_This is a suggested implementation which can be used alongside the Secrets Store CSI Driver._
+
+When [auto rotation of secrets](./secret-auto-rotation.md) is enabled, workloads which depend upon secrets will need to either 
+- watch for updates to secrets and reload these in their runtime, or
+- be restarted to pick up the latest secrets when they change
+
+A solution such as [Reloader](https://github.com/stakater/Reloader) could be used to watch for updates to Kubernetes Secrets or ConfigMaps and restart pods when a change is detected. However, if secret values are mounted as volumes in the pods, that solution is not suitable.
+
+Using custom resources created by the Secrets Store CSI Driver, a Kubernetes Controller can be used to detect when secrets are updated by the driver and to restart affected pods.
+
+## SecretProviderClassPodStatus custom resource
+
+The relevant custom resource is [`SecretProviderClassPodStatus`](https://secrets-store-csi-driver.sigs.k8s.io/concepts#secretproviderclasspodstatus).
+
+Each `SecretProviderClassPodStatus` CR has a one-to-one relationship with a pod which references secrets using the Secrets Store CSI Driver. The CR identifies the pod name, namespace and other attributes. The driver manages the lifecyle of `SecretProviderClassPodStatus` which is linked to the lifecycle of the affected pod.
+
+```mermaid
+stateDiagram-v2
+    state "SecretProviderClassPodStatus\nGeneration: 1" as g1
+    state "SecretProviderClassPodStatus\nGeneration: n" as gn
+
+    [*] --> g1: pod create with secret from csi
+    g1 --> gn: secret updated in pod
+    gn --> [*]: pod restart
+```
+
+When the driver sets a secret value for a new pod, a `SecretProviderClassPodStatus` CR is created with the `Generation` attribute set to `1`. 
+
+Whenever the driver updates the secret value, the value of the `Generation` attribute is incremented.
+
+If a pod is restarted, the CR is deleted and a new CR created with `Generation: 1`.
+
+`SecretProviderClassPodStatus` CRs persist across lifetimes of the secrets-store-csi-driver.
+
+## Outline of Controller function
+
+1. Reconcile
+
+    The controller reconciles instances of the `SecretProviderClassPodStatus` CR and deletes (to restart) the associated pod if required.
+
+    If a `SecretProviderClassPodStatus` has `Generation: 1`, it is linked to a newly created pod. The pod should not be restarted.
+
+    If a `SecretProviderClassPodStatus` has `Generation` > 1, it is linked to a pod in which the secrets-store-csi-driver has updated a secret. The pod should be restarted (if it has opted-in for automatic restarting).
+
+1. Rolling restart
+
+    On reconciling a pod which should be updated, check `metadata.ownerReferences` and walk up to a Deployment (or similar) if present.
+
+    If a `Deployment` is found:
+
+    - Do not restart pod
+    - Update the Deployment to trigger a rolling restart
+    - If the number of replicas > 1, update the Deployment once only
+
+    To restart a deployment, the controller sets a timestamped annotation in the deployment
+
+    ```
+      template:
+        metadata:
+          annotations:
+            my.controller/restartedAt: "2024-09-05T14:06:29Z"
+    ```
+
+    Else: delete pod.
+
+1. Opt-in to automatic pod restarting
+
+    Automatic restarting of pods when secrets are updated could be an opt-in behaviour. Unless the pod declares its opt-in, it should not be restarted by the controller.
+
+    The opt-in could be indicated via an optional annotation set on the pod:
+    ```
+      kind: pod
+      metadata:
+        annotations:
+          my.controller/restartOnChange: true
+    ```
+
+## Implementation notes
+
+The [operator-sdk](https://github.com/operator-framework/operator-sdk) can be used to scaffold an implementation project.
+
+1. Scaffolding the project
+
+    ```
+    operator-sdk init --repo=<your repo>
+    operator-sdk create api --version v1alpha1 --kind SecretProviderClassPodStatus --resource=false --controller=true
+    ```
+
+1. Custom resources
+
+    The controller does not manage custom resources of its own. It simply watches a custom resource provided by the Secrets Store CSI Driver.
+
+1. Permissions required
+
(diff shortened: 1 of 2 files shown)


ewan-chalmers:
Looking at https://deploy-preview-1648--kubernetes-sigs-secrets-store-csi-driver.netlify.app/topics/auto-restart-on-rotation, I note that mermaid source is not rendered as a diagram in the mdbook output ([as it is on github](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/4d17d7ac66c809f63aa7e9a006343c019c230624/docs/book/src/topics/auto-restart-on-rotation.md)).

Simplest thing is to remove the doc, but I could also look into whether mdbook configuration can be persuaded to render mermaid diagrams.

ewan-chalmers:
Well... that took a few more tries than I expected. The diagram now renders (FWIW - not a very exciting diagram) - https://deploy-preview-1648--kubernetes-sigs-secrets-store-csi-driver.netlify.app/topics/auto-restart-on-rotation#secretproviderclasspodstatus-custom-resource

While doing this, I notice that we should probably upgrade mbook and mbook-toc versions (as well as this new mbook-mermaid). I had to go quite a way back in mbook-mermaid tags to get one that works with our mbook version.

I'm happy to extract mermaid enabling into a separate PR or remove it altogether (along with the diagram src in this PR) if preferred.

ewan-chalmers:
I'm unfamiliar with squashing commits and looks like there are loads of different ways to do so :/

docs/book/src/topics/auto-restart-on-rotation.md
```diff
@@ -0,0 +1,120 @@
+# Auto restart of pods when secret is rotated
+
+_This is a suggested implementation which can be used alongside the Secrets Store CSI Driver._
+
+When [auto rotation of secrets](./secret-auto-rotation.md) is enabled, workloads which depend upon secrets will need to either 
+- watch for updates to secrets and reload these in their runtime, or
+- be restarted to pick up the latest secrets when they change
+
+A solution such as [Reloader](https://github.com/stakater/Reloader) could be used to watch for updates to Kubernetes Secrets or ConfigMaps and restart pods when a change is detected. However, if secret values are mounted as volumes in the pods, that solution is not suitable.
```
benjaminapetersen, line 9: From an official SIG-Auth project, should we be linking out to this in the formal docs?

benjaminapetersen: This seems in tone a bit like a blog post for solving a particular problem, rather than a docs site entry. 🤔

benjaminapetersen: Per our discussion in the sync meeting, perhaps we keep, but just note that this is not supported or tested.

ewan-chalmers: You could see this doc as a continuation from https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/docs/book/src/topics/secret-auto-rotation.md.

I aimed for similar tone.

That doc links to Reloader.

I'm open to suggestions.

My motivation here is that the solution I arrived at and implemented was
- not obvious based on existing resources 
- fairly straightforward given the necessary pointers

ewan-chalmers: @benjaminapetersen I have added a note to the top of the doc. If this is not what you have in mind, could you suggest alternative? Thanks

ewan-chalmers:
@aramase @ritazh Any suggestion on how I could move this PR to done?

docs/book/src/topics/auto-restart-on-rotation.md
```diff
@@ -0,0 +1,120 @@
+# Auto restart of pods when secret is rotated
+
+> NOTE: This is a suggested implementation which can be used alongside the Secrets Store CSI Driver. The solution is neither supported nor tested by the Secrets Store CSI Driver project.
+
+When [auto rotation of secrets](./secret-auto-rotation.md) is enabled, workloads which depend upon secrets will need to either 
+- watch for updates to secrets and reload these in their runtime, or
+- be restarted to pick up the latest secrets when they change
+
+A solution such as [Reloader](https://github.com/stakater/Reloader) could be used to watch for updates to Kubernetes Secrets or ConfigMaps and restart pods when a change is detected. However, if secret values are mounted as volumes in the pods, that solution is not suitable.
```
aramase, line 9:
```suggestion
A solution like [Reloader](https://github.com/stakater/Reloader) can watch updates to Kubernetes Secrets or ConfigMaps and restart pods when a change is detected. However, if secret values are mounted as volumes in the pods, that solution is not suitable.
```

docs/book/src/topics/auto-restart-on-rotation.md
```diff
@@ -0,0 +1,120 @@
...
+
+Using custom resources created by the Secrets Store CSI Driver, a Kubernetes Controller can be used to detect when secrets are updated by the driver and to restart affected pods.
```
aramase, line 11:
```suggestion
Using custom resources created by the Secrets Store CSI Driver, a Kubernetes controller can detect when secrets are updated by the driver and restart the associated pods.
```

docs/book/src/topics/auto-restart-on-rotation.md
```diff
@@ -0,0 +1,120 @@
...
+
+## SecretProviderClassPodStatus custom resource
+
+The relevant custom resource is [`SecretProviderClassPodStatus`](https://secrets-store-csi-driver.sigs.k8s.io/concepts#secretproviderclasspodstatus).
```
aramase, line 15:
```suggestion
The relevant custom resource is [`SecretProviderClassPodStatus`](./concepts.md#secretproviderclasspodstatus).
```

something like this, I haven't tested it, so please confirm it works.

docs/book/src/topics/auto-restart-on-rotation.md
```diff
@@ -0,0 +1,120 @@
...
+
+Each `SecretProviderClassPodStatus` CR has a one-to-one relationship with a pod which references secrets using the Secrets Store CSI Driver. The CR identifies the pod name, namespace and other attributes. The driver manages the lifecyle of `SecretProviderClassPodStatus` which is linked to the lifecycle of the affected pod.
```
aramase, line 17:
```suggestion
Each `SecretProviderClassPodStatus` custom resource (CR) has a one-to-one relationship with a pod that references secrets using the Secrets Store CSI Driver. The CR includes the pod name, namespace and other attributes. The driver manages the lifecycle of `SecretProviderClassPodStatus` which is tied to the lifecycle of the associated pod.
```

aramase, line 17: It's 1:1 mapping for pod and SPC, so clarify that here

aramase:
/retitle docs: Document auto-restart of pods on secret rotation

docs/book/src/topics/auto-restart-on-rotation.md
ewan-chalmers: You have, or I should?

docs/book/src/topics/auto-restart-on-rotation.md
ewan-chalmers: committed working link

ewan-chalmers:
> Reading through the implementation details, I believe the controller restart will now generate twice the number of calls for the associated pod. The `SecretProviderClassPodStatus` is only updated by the driver after it has updated the mounted content with the new secrets. By this point, we’ve already made 𝑂(number of secrets) calls to the external secrets store. When the controller restarts the pod, it will trigger a new mount operation, resulting in another 𝑂(number of secrets) calls. In my opinion, this essentially defeats the purpose of auto rotation.

I may be misunderstanding, but I think you are pointing to a potentially undesirable consequence of this implementation rather than to something which defeats the purpose. The purpose is: if mounted secrets are updated, pod will be restarted; pod code does not need to handle dynamic secret updating. This does work. However, are you pointing to the fact that for one secret update operation in the secrets store, the driver will now read the secrets from there and create mount for the pod twice - because it does this again after controller restarts the pod? If so, do you agree it is not 'defeating the purpose' but rather a consequence which might be undesirable and so should be mentioned?

ewan-chalmers:
Thanks for review @aramase

docs/book/src/topics/auto-restart-on-rotation.md
ewan-chalmers: Done

maxzhang2017:
Hi, @ewan-chalmers 
I found if auto rotation of secrets is enabled, the Generation attribute of SecretProviderClassPodStatus is increasing all the time, regardless of whether the secret data has changed or not.

PR #1676, opened 2024-10-25
fix: update the hash generation logic

aramase:
/kind bug

Fix the hash generation logic to prevent hash collision with different maps as input. This can be caused because of the concatenation logic.

The test added in https://github.com/kubernetes-sigs/secrets-store-csi-driver/commit/b94492d3eeda05e1589e1f4880ef2a5bb02f35a0 shows the bug in the current logic.
```bash
=== RUN   TestGenerateSHAFromSecret/different_keys_with_the_same_concatenated_result_but_should_not_match
    secret_test.go:434: 
        	Error Trace:	/Users/anishramasekar/go/src/sigs.k8s.io/secrets-store-csi-driver/pkg/util/secretutil/secret_test.go:434
        	Error:      	Not equal: 
        	            	expected: false
        	            	actual  : true
        	Test:       	TestGenerateSHAFromSecret/different_keys_with_the_same_concatenated_result_but_should_not_match
--- FAIL: TestGenerateSHAFromSecret/different_keys_with_the_same_concatenated_result_but_should_not_match (0.00s)
```

/assign enj

### DIFF
--- a/pkg/util/secretutil/secret.go
+++ b/pkg/util/secretutil/secret.go
@@ -23,15 +23,16 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
-	"io"
+	"math"
 	"os"
-	"sort"
 	"strings"
 
 	secretsstorev1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
 
+	"golang.org/x/crypto/cryptobyte"
 	"golang.org/x/crypto/pkcs12"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/util/secretutil/secret.go
```diff
@@ -217,23 +217,32 @@ func GetSecretData(secretObjData []*secretsstorev1.SecretObjectData, secretType
...
+	}
+
+	b := cryptobyte.NewBuilder(nil)
+	b.AddUint32(uint32(len(data)))
+
+	keys := make([]string, 0, len(data))
+	for k := range data {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
enj, line 231:
```suggestion
	keys := sets.StringKeySet(data).List()
```

enj (commented):
Minor comment.

Issue #1437, opened 2024-02-12
Feat: Add support to define (overwrite) SecurityContext


cbugneac-nex:
**Describe the solution you'd like**
One of requirements from our Cyber security team is to drop capabilities where possible to improve security posture (reduce attack surface) for containers running in k8s.

In values.yaml file to define something like (example):
```yaml
...
securityContext:
  capabilities:
    drop:
      - ALL
  runAsNonRoot: true
  allowPrivilegeEscalation: false
  seccompProfile:
    type: RuntimeDefault
```

**Anything else you would like to add:**
At the moment, this is not possible as it's hardcoded https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/templates/secrets-store-csi-driver.yaml#L113

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): v1.4.1
- Kubernetes version: (use `kubectl version`): EKS version 1.26.12

PR #1438, opened 2024-02-12
Add support to define (overwrite) of securityContext or pod security context

cbugneac-nex:
**What this PR does / why we need it**:
Add support to define (overwrite) SecurityContext for containers & pod security context.

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #1437

**Special notes for your reviewer**:

Link helm chart:
```bash 
$ helm lint ./
==> Linting ./

1 chart(s) linted, 0 chart(s) failed
```

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

PR #1733, opened 2025-01-17
Feature: Support security context (#1437)

max-allan:
**What type of PR is this?**

/kind feature

Fixes https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/1437

Fixed version of https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1438/files

Moves 'privileged: true' into 'linux.securityContext` and does not change the `podSecurityContext` (but there is a value for it, if people want to do that).

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [n/a] adds unit tests


manifest_staging/charts/secrets-store-csi-driver/values.yaml
```diff
@@ -22,6 +25,9 @@ linux:
       #   memory: "512Mi"
     annotations: {}
     podLabels: {}
+  # The secrets-store container security context
+  securityContext:
+    privileged: true
```
aramase, line 30: `privileged: true` is **required** for the driver to function correctly. Moving this to `values.yaml` could cause the driver functionality to fail if not set correctly by the user.

`privileged: true` should instead remain in the manifest and this field should append to the existing `securityContext`.

aramase, line 30: `privileged: true` is **required** for the driver to function correctly. Moving this to `values.yaml` could cause the driver functionality to fail if not set correctly by the user.

`privileged: true` should instead remain in the manifest and this field should append to the existing `securityContext`.

aramase (changes requested):

manifest_staging/charts/secrets-store-csi-driver/values.yaml
max-allan: The whole point of the original request is that some people's security policy is averse to `privileged: true`

If the user chooses to disable it, then we must assume they have good reason for doing so and understand the consequences and will tune the exact CAP_ capabilities they need to grant. (Probably CAP_SYS_ADMIN).

Mixing a hidden privileged setting with a Context will only cause more confusion. "I set only ... but it runs with privileged: true as well."

PR #1749, opened 2025-02-10
Removed TokenClient and use CSI token propagation

### LINKED ISSUE #585: Using `RequiresRepublish` for rotation

micahhausler:
**What type of PR is this?**

/kind feature
/kind cleanup

**Which issue(s) this PR fixes**:
Fixes #585

* I made the remount behavior dependent on whether `requiresRequblish` is set in the CSI driver. The default is `false`, which preserves existing behavior and won't issue new `Mount()` calls to plugins.
* The driver already had RBAC permissions to get/list/watch its own CSI Driver resource by name, so no new K8s permissions are needed.
* Given that 1.24 has been EOL for years, I thought it'd be safe to drop the TokenClient everywhere.
* By removing the need to create any service account token on all nodes, this is a _huge_ security win: the driver will only propagate the token provided by Kubelet on to the plugins. 

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests


enj:
@micahhausler is this trying to redo #1622 which is almost complete or are they separate?

micahhausler:
I hadn't seen #1622 and wanted an excuse to try Cursor. I'm happy to close this and review #1622

aramase:
Closing this in favor of https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1622.

/close

PR #1755, opened 2025-02-17
feat: Add VolumeID to Mount request

### LINKED ISSUE #1754: Supply VolumeID in Mount request attributes

Once RequiresRepublish is implemented (#585), a ServiceAccount token will be supplied every ~75 seconds by the Kubelet for each volume, even though the underlying token won't change until about ~80% (with jitter) of the SA token's lifetime. (Default min/default token lifetime is 10m/24h, so token rotation is ~8m/19h12m)

This could trigger an unnecessary call volume for providers that might today synchronously exchange a token  for cloud credentials that assume they'll only ever be called on initial setup, but don't need to refresh exchanged credentials with the same frequency as the repeated NodePublishVolume calls by the Kubelet. 

This naturally will lead providers to likely cache a provided token and cloud credentials, but this leads to the problem of what to use for a cache key.

The [current Mount info supplied to providers](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v1.4.8/pkg/secrets-store/nodeserver.go#L118-L183) includes the Pod name/namespace/uid, which is not necessarily enough for volume uniqueness (ex: pod name can be reused across time (although UID won't be reused), and pods can have multiple volumes: potentially using multiple SecretProviderClasses) to store a cached token and/or credential. 

While its true that the token itself could be used as a cache key by providers, its probably not a good idea as its much more likely to be logged, and possible cloud credentials exchanged for a given token could differ between volume mounts (ex: pod1/mount1 uses token1 for credentials in cloud region us-east-1, pod1/mount2 uses the same token1 for credentials in cloud region us-west-2. The AWS provider allows such a configuration)

To give providers a way to uniquely cache tokens and exchanged cloud credentials, we should supply the VolumeID from the NodePubilshVolume request into the Mount request passed on to providers.

N/A

**Environment:**

- Secrets Store CSI Driver version: (use the image tag): Any
- Kubernetes version: (use `kubectl version`): Any modern version

micahhausler:
**What type of PR is this?**

/kind feature

This PR adds the VolumeID to the request so providers can make smarter caching decisions once `requiresRepublish` is implemented.

Fixes #1754

I opted to name the key with the DNS namespace for this project as there isn't a corresponding CSI key I could find in the Kubernetes codebase (the volume ID is just an attribute of the request, not an encoded key). Let me know if there is a better key name to use

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/secrets-store/nodeserver.go
+++ b/pkg/secrets-store/nodeserver.go
@@ -65,6 +65,8 @@ const (
 	// CSIPodServiceAccountTokens is the service account tokens of the pod that the mount is created for
 	CSIPodServiceAccountTokens = "csi.storage.k8s.io/serviceAccount.tokens" //nolint
 
+	SecretStoreVolumeID = "secrets-store-csi-driver.sigs.k8s.io/volume.id"
+
 	secretProviderClassField = "secretProviderClass"
 )
 
@@ -181,6 +183,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	for k, v := range attrib {
 		parameters[k] = v
 	}
+	// Add the volume ID to the parameters
+	parameters[SecretStoreVolumeID] = volumeID
 	// csi.storage.k8s.io/serviceAccount.tokens is empty for Kubernetes version < 1.20.
 	// For 1.20+, if tokenRequests is set in the CSI driver spec, kubelet will generate
 	// a token for the pod and send it to the CSI driver.
--- a/docs/book/src/providers.md
+++ b/docs/book/src/providers.md
@@ -35,6 +35,18 @@ The driver uses gRPC to communicate with the provider. To implement a secrets-st
 
 See [design doc](https://docs.google.com/document/d/10-RHUJGM0oMN88AZNxjOmGz0NsWAvOYrWUEV-FbLWyw/edit?usp=sharing) for more details.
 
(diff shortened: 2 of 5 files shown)


benjaminapetersen (commented):
Seems reasonable to me.
@aramase or @enj for 👀.

pkg/secrets-store/nodeserver.go
```diff
@@ -65,6 +65,8 @@ const (
 	// CSIPodServiceAccountTokens is the service account tokens of the pod that the mount is created for
 	CSIPodServiceAccountTokens = "csi.storage.k8s.io/serviceAccount.tokens" //nolint
 
+	SecretStoreVolumeID = "secrets-store-csi-driver.sigs.k8s.io/volume.id"
```
benjaminapetersen, line 68: Should have a test to prove this is present?

micahhausler: I can add a test and doc note about it

micahhausler: Turns out there is already a test ensuring this is present. I'll just add teh docs

aramase (commented):
The changes look fine!

This can only be merged after we support `RequiresRepublish` because the rotation controller doesn't have access to the volumeID.

/hold

> token could differ between volume mounts (ex: pod1/mount1 uses token1 for credentials in cloud region us-east-1, pod1/mount2 uses the same token1 for credentials in cloud region us-west-2. The AWS provider allows such a configuration)

Isn't it easier for the provider to use the region in this case or other provider specific information another dimension of cache key? The pod uid will be the first dimension.

aramase (commented):
@micahhausler is this change still required? see https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/1755#pullrequestreview-2773318641.

If it's still required, please rebase the PR when you get a chance.

aramase:
/close

due to inactivity, feel free to reopen when ready for review.

PR #1777, opened 2025-03-25
Aramase/s/fix cves

aramase:
/kind cleanup

- Bump packages to fix CVEs
- Disable trivy packge test for driver-crds because we need a new version of `kubectl` binary and the current failures can't be resolved until then.


pkg/util/secretutil/secret_test.go
```diff
@@ -101,6 +101,7 @@ koN/frqYab5Ek3kauj1iqW7rPkrFCqT2evh0YRqb1bFsCLJrRNxnOZ5wKXV/OYQa
 QX5t0wFGCZ0KlbXDiw==
 -----END CERTIFICATE-----
 `
+	//nolint:gosec
```
enj, line 104: Add more details

Makefile
```diff
@@ -164,7 +164,7 @@ image-scan: $(TRIVY) ## Run image-scan
 	$(TRIVY) image --severity MEDIUM,HIGH,CRITICAL $(CRD_IMAGE_TAG)
 	# show vulnerabilities that have been fixed
 	$(TRIVY) image --exit-code 1 --ignore-unfixed --severity MEDIUM,HIGH,CRITICAL $(IMAGE_TAG)
-	$(TRIVY) image --exit-code 1 --ignore-unfixed --severity MEDIUM,HIGH,CRITICAL $(CRD_IMAGE_TAG)
+	$(TRIVY) image --vuln-type os  --exit-code 1 --ignore-unfixed --severity MEDIUM,HIGH,CRITICAL $(CRD_IMAGE_TAG)
```
enj, line 167: What does this mean?

enj (commented):
Minor comments.

Makefile
aramase: This is to only scan and fail for CVEs in the base image for `driver-crds` which contains the kubectl binary.

PR #2059, opened 2026-08-09
fix: skip no-op secret updates in createOrUpdateK8sSecret

### LINKED ISSUE #2057: 1.6.0: Network and CPU throughput increased after upgrade.
# v1.6.0: syncSecret rewrites every managed Secret on every reconcile, causing sustained API server write load and driver CPU/network increase

After upgrading the driver from v1.5.6 to v1.6.0 (Helm chart 1.5.6 → 1.6.0), driver CPU rose ~4x and driver pod network receive rose by 10x
`secretRotation` are using their default values in the helm chart. We do believe that this code entry might be the cause.

v1.5.6 — [`createK8sSecret`](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v1.5.6/controllers/secretproviderclasspodstatus_controller.go#L415-L435) returned early when the Secret already existed:

```go
err := r.writer.Create(ctx, secret)
if err == nil { ...; return nil }
if apierrors.IsAlreadyExists(err) {
    return nil                       // no-op
}
```

v1.6.0 — [`createOrUpdateK8sSecret`](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v1.6.0/controllers/secretproviderclasspodstatus_controller.go#L401-L428) follows the failed `Create` with a full `Update`, with no comparison against the stored object:

```go
err := r.writer.Create(ctx, secret)
if err == nil { ...; return nil }
if !apierrors.IsAlreadyExists(err) {
    return err
}
klog.V(5).InfoS("Kubernetes secret is already created", ...)
err = r.writer.Update(ctx, secret)   // unconditional, even when data is identical
```

So every `SecretProviderClassPodStatus` reconcile now costs two write requests per synced Secret: a `Create` that fails with `AlreadyExists`, then a full-body `Update`. 

I verified the writes are no-ops: for the same Secret, `.data` is byte-identical across consecutive updates and only `resourceVersion` advances.

**Measurements**

Two production-shaped clusters with the same workload type, one on each version. Note the v1.5.6 cluster is the *larger* of the two:

| | cluster A — v1.6.0 | cluster B — v1.5.6 |
|---|---|---|
| nodes | 43 | 77 |
| `SecretProviderClassPodStatus` objects | 5,972 | **14,239** |
| Secrets labelled `secrets-store.csi.k8s.io/managed=true` | 7,273 | 13,965 |
| driver container CPU (`kubectl top`, per pod) | **15m avg / 37m peak** | 7m avg / 9m peak |
| `apiserver_request_total{resource="secrets",verb="POST"}` | **~29/s** (494 in 17s) | **2 total, process lifetime** |
| `apiserver_request_total{resource="secrets",verb="PUT"}` | **~29/s** (495 in 17s) | metric absent |
| driver pod `container_network_receive_bytes_total` | **90–102 KB/s** | 44 B/s |
| driver pod `container_network_transmit_bytes_total` | 5.6–7.8 KB/s | 35 B/s |

The POST and PUT rates track each other almost exactly (494 vs 495 over the same interval), which is the failed-`Create`/`Update` pair, once per reconcile.

**Network amplification**

The Secret informer is filtered by label only, not by node ([`main.go`](https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/v1.6.0/cmd/secrets-store-csi-driver/main.go)):

```go
&corev1.Secret{}: {
    Label: labels.SelectorFromSet(labels.Set{controllers.SecretManagedLabel: "true"}),
},
```

Every driver pod therefore watches all managed Secrets cluster-wide and receives every write any other node makes. At ~29 writes/s and ~3 KB per Secret object (metadata + `managedFields`) that predicts ~87 KB/s per pod, which matches the 90–102 KB/s measured. Across 41 driver pods that is roughly **3.9 MB/s (~31 Mbps) of API server egress** with no state change behind it.

This selector is identical in v1.5.6, so the watch itself is not the regression — v1.5.6 simply produced almost no Secret writes for it to deliver. One pod in cluster A showed only 179 B/s; it sits on a node with zero `SecretProviderClassPodStatus` objects, so no reconcile ever ran and controller-runtime never lazily started its Secret informer. That asymmetry is consistent with the fan-out being the source of the traffic.

A reconcile of an unchanged `SecretProviderClassPodStatus` should not write to the API server. If the synced Secret already exists with the desired content, the driver should leave it alone, as v1.5.6 did.

Updating on drift is a reasonable improvement over v1.5.6's create-once behaviour — the problem is that the update is unconditional. Comparing the desired `data`/`type`/labels/annotations against the existing object and only calling `Update` when they differ would keep the new behaviour and remove the steady-state write load.

Also worth noting: each `Update` re-encrypts the Secret at rest, so on managed clusters with KMS envelope encryption this write load reaches etcd and the KMS path, not just the driver.

Impact scales with `managed Secrets × nodes`, so larger clusters are hit harder. Extrapolating cluster B's inventory (14k Secrets, 77 nodes) to v1.6.0 puts it around 60 writes/s fanned out to ~77 watchers, which is why we have held the upgrade at one non-production cluster.

AWS provider (`secrets-store-csi-driver-provider-aws` 3.0.0). The provider is idle throughout (1m CPU, 10–13 MiB), and provider calls are not involved — the load is entirely between the driver and the API server. The behaviour is in driver code and looks provider-independent.

**Environment:**

- Secrets Store CSI Driver version: `registry.k8s.io/csi-secrets-store/driver:v1.6.0` (regression vs `v1.5.6`)
- Kubernetes version: v1.35.6 (EKS) — comparison cluster on v1.34.9 (EKS)
- Helm chart: `secrets-store-csi-driver` 1.6.0, `syncSecret.enabled=true`, `enableSecretRotation` unset, `requiresRepublish: false`
- Mode: Linux

pujitha24:
**What type of PR is this?**

/kind bug

Since v1.6.0, `createOrUpdateK8sSecret` follows a failed `Create` (`AlreadyExists`) with an unconditional `Update`, even when the secret's data/type/labels/annotations already match what's stored. This means every reconcile of every synced `SecretProviderClassPodStatus` writes to the API server, regardless of whether anything actually changed, causing sustained write load, unnecessary etcd/KMS re-encryption, and (via the cluster-wide managed-Secret informer) network fan-out to every driver pod.

This PR fetches the existing secret before falling back to `Update` and skips the write when the desired type/data/labels/annotations already match, restoring the effective no-op behavior of v1.5.6 while keeping the v1.6.0 improvement of updating on real drift.

Fixes #

- The existing-secret read uses `r.reader` (the controller-runtime manager cache), which is already populated by the managed-Secret informer this code path relies on, so the comparison is a cache read rather than an extra API call. If the cache hasn't yet synced a very recent out-of-band change to the secret, a reconcile could skip an update it should perform; this self-heals on the next reconcile/resync and is not a correctness issue.
- The comparison uses `k8s.io/apimachinery/pkg/api/equality`'s `Semantic.DeepEqual` rather than `reflect.DeepEqual`, because `Reconcile` builds `annotationsMap` as a fresh empty (non-nil) map whenever the `SecretProviderClass` has no explicit annotations, while a fetched secret's `Annotations` comes back `nil` for that case. `reflect.DeepEqual` treats nil vs. empty maps as unequal, which would have defeated the fix for the common case of secrets with no explicit annotations; `Semantic.DeepEqual` correctly treats them as equal.
- If the secret is missing from the cache (e.g. a very recent create not yet observed), the code falls back to the prior unconditional-`Update` behavior rather than erroring.
- This does not touch `patchSecretWithOwnerRef`, which is a separate, independently-scheduled code path (owner-ref patcher) unaffected by this change.

**Validation**:

- `go build ./...` succeeds.
- `go vet ./controllers/...` succeeds.
- `go test ./controllers/... -v` passes, including two new targeted regression tests:
  - `TestCreateOrUpdateK8sSecretNoopWhenUnchanged`: reconciling with identical data/type/labels/annotations does not bump the secret's `ResourceVersion` (i.e. no `Update` call is issued); reconciling with different data does bump it and persists the new data.
  - `TestCreateOrUpdateK8sSecretNoopWhenAnnotationsUnset`: mirrors `Reconcile`'s exact construction of an empty (non-nil) `annotationsMap` against a stored secret with `Annotations == nil`, and asserts no update occurs.
- Both new tests were confirmed to fail against the pre-fix code (unconditional `Update`) with a `ResourceVersion` mismatch, and to fail again if the map comparison is reverted to `reflect.DeepEqual` instead of `apiequality.Semantic.DeepEqual` — i.e. each test is a genuine regression test for the specific defect it targets, not a vacuous pass.
- No live cluster / e2e reproduction of the reported API-server write-rate increase was run in this environment; the driver's own `docs/book/src/testing.md` documents unit tests via `make test` as the local validation path and states e2e runs automatically on Prow for PRs, without requiring a local live-cluster run for a change of this kind. The fix is validated as a targeted, provable code defect (unconditional write on unchanged state) via failing-then-passing unit tests, consistent with the root cause described in the linked issue.

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

Fixes #2057

```release-note
fix: skip no-op secret updates in createOrUpdateK8sSecret
```

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/secrets-store-csi-driver/pkg/util/secretutil"
 
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
@@ -420,12 +421,36 @@ func (r *SecretProviderClassPodStatusReconciler) createOrUpdateK8sSecret(ctx con
 	}
 
 	klog.V(5).InfoS("Kubernetes secret is already created", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
+
+	existing := &corev1.Secret{}
+	if err := r.reader.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, existing); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	} else if secretUpToDate(existing, secret) {
+		klog.V(5).InfoS("Kubernetes secret already up to date, skipping update", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
+		return nil
+	}
+
 	err = r.writer.Update(ctx, secret)
 	if err != nil {
 		return err
 	}
 	klog.V(5).InfoS("successfully updated Kubernetes secret", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
 	return nil
+}
+
+// secretUpToDate returns true if the existing secret already has the desired
+// type, data, labels and annotations, so writing it again would be a no-op.
+// apiequality.Semantic.DeepEqual is used instead of reflect.DeepEqual so that
+// a nil map (as returned by the API server for an unset field) and an empty
+// map (as built by the caller when no labels/annotations are configured) are
+// treated as equal.
+func secretUpToDate(existing, desired *corev1.Secret) bool {
+	return existing.Type == desired.Type &&
+		apiequality.Semantic.DeepEqual(existing.Data, desired.Data) &&
+		apiequality.Semantic.DeepEqual(existing.Labels, desired.Labels) &&
+		apiequality.Semantic.DeepEqual(existing.Annotations, desired.Annotations)
 }
 
 // patchSecretWithOwnerRef patches the secret owner reference with the spc pod status
--- a/controllers/secretproviderclasspodstatus_controller_test.go
+++ b/controllers/secretproviderclasspodstatus_controller_test.go
@@ -188,6 +188,76 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
 	g.Expect(secret.Name).To(Equal("my-secret2"))
 }
 
+func TestCreateOrUpdateK8sSecretNoopWhenUnchanged(t *testing.T) {
+	g := NewWithT(t)
+
+	scheme, err := setupScheme()
+	g.Expect(err).NotTo(HaveOccurred())
+
+	labels := map[string]string{"environment": "test"}
+	annotations := map[string]string{"kubed.appscode.com/sync": "app=test"}
+	datamap := map[string][]byte{"key": []byte("value")}
+
+	existing := newSecret("my-secret", "default", labels, annotations)
+	existing.Type = corev1.SecretTypeOpaque
+	existing.Data = datamap
+
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(existing).Build()
+	reconciler := newReconciler(client, scheme, "node1")
+
+	before := &corev1.Secret{}
+	g.Expect(client.Get(context.TODO(), types.NamespacedName{Name: "my-secret", Namespace: "default"}, before)).To(Succeed())
+
+	// reconciling with the exact same data/type/labels/annotations should not issue an update.
+	err = reconciler.createOrUpdateK8sSecret(context.TODO(), "my-secret", "default", datamap, labels, annotations, corev1.SecretTypeOpaque)
+	g.Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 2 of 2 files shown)


pujitha24:
/assign @enj

This has been green and rebased for a week — happy to adjust anything that would make review easier.

stlaz:
/close
superseded by https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/2065

PR #2060, opened 2026-08-09
fix: make SecretProviderClassPodStatus reconciliation idempotent

### LINKED ISSUE #2058: SecretProviderClassPodStatus reconciliation unconditionally creates and updates existing objects

In v1.6.0, every call to `createOrUpdateSecretProviderClassPodStatus` attempts to create the `SecretProviderClassPodStatus` object first. For an object that already exists, the normal steady-state sequence is:

1. `POST` the status object and receive `409 AlreadyExists`.
2. `GET` the existing object.
3. Unconditionally `PUT` the object, even when its labels, status, and owner references already match the desired state.

The relevant flow is in `pkg/secrets-store/utils.go`:

```go
if err = c.Create(ctx, spcPodStatus); err == nil || !apierrors.IsAlreadyExists(err) {
    return err
}

spcps := &secretsstorev1.SecretProviderClassPodStatus{}
if err = c.Get(ctx, client.ObjectKey{Name: spcpsName, Namespace: namespace}, spcps); err != nil {
    if !apierrors.IsNotFound(err) {
        return err
    }
    if err = reader.Get(ctx, client.ObjectKey{Name: spcpsName, Namespace: namespace}, spcps); err != nil {
        return err
    }
}

spcps.Labels[secretsstorev1.InternalNodeLabel] = nodeID
spcps.Status = spcPodStatus.Status
spcps.OwnerReferences = spcPodStatus.OwnerReferences

return c.Update(ctx, spcps)
```

This makes an expected `AlreadyExists` response and a full update part of every repeated reconciliation for an existing object. In a large cluster, repeated `NodePublishVolume` or republish activity can therefore create sustained API server traffic even when the mounted state has not changed.

A generic reproduction is:

1. Mount a `SecretProviderClass` in a pod so its `SecretProviderClassPodStatus` exists.
2. Cause the driver to process the same volume again without changing the pod, mounted objects, versions, target path, or node assignment.
3. Observe a `POST` returning `409`, followed by a `PUT` for the same status object.
4. Repeat the operation and observe the same request pair despite no semantic state change.

Reconciliation should be idempotent. If the existing `SecretProviderClassPodStatus` already matches the desired labels, status, and owner references, the driver should make no mutating API request.

Suggested implementation:

1. Build the desired object as today.
2. Read the existing object first. Because the cache can exclude an object carrying another node label, retain the direct-reader fallback before concluding that it is absent.
3. Create only when an authoritative read confirms the object does not exist.
4. Compare the existing and desired node label, `Status`, and `OwnerReferences`; return without writing when they are semantically equal.
5. Update only when one of those fields differs.
6. Use a bounded retry around genuine create/update races (`AlreadyExists` after an absent read and `Conflict` during update), fetching a fresh object on each attempt.

Suggested tests:

- Missing object: one create, no update.
- Existing unchanged object: no create and no update.
- Existing changed object: one update.
- Update conflict: bounded retry with a fresh read.
- Create race: recover from `AlreadyExists` and converge without an unconditional extra write.
- Cache miss caused by a different internal node label: direct-reader lookup still finds and updates the object.

This appears similar in shape to #2057, which reports unconditional create/update behavior for synced Kubernetes Secrets in v1.6.0, but this issue concerns the separate `SecretProviderClassPodStatus` write path.

`RetryOnConflict` by itself would not address the dominant behavior because `AlreadyExists` is expected for every existing object and the current update is unconditional. The primary fix is read/compare/write idempotency; bounded retries should only handle real races.

It may also be useful to examine why unchanged status reconciliation is invoked repeatedly, but making this helper idempotent prevents those invocations from becoming API mutations regardless of their source.

Provider-independent. This behavior is in the core driver status-object helper and does not depend on a provider implementation.

**Environment:**

- Secrets Store CSI Driver version: `v1.6.0`
- Kubernetes version: Provider- and version-independent code path; no specific Kubernetes version is required to demonstrate the request sequence.

pujitha24:
**What type of PR is this?**

/kind bug

`createOrUpdateSecretProviderClassPodStatus` unconditionally attempted a `Create` on every
reconcile, and once the object already existed, unconditionally ran a `Get` followed by a `Put`
(`Update`) even when the label, `Status`, and `OwnerReferences` already matched the desired state.
In steady state (repeated `NodePublishVolume`/republish activity for an unchanged mount), this
produces a `POST` returning `409 AlreadyExists` followed by an unconditional `PUT` on every
reconcile, adding unnecessary write traffic to the API server without changing any observable
behavior.

This PR makes the helper idempotent: it reads the existing object first (falling back to a direct
API read if the cache doesn't have it, since it may be labeled for a different node and therefore
excluded from this node's cache), compares the node label, `Status`, and `OwnerReferences` against
the desired values, and returns without writing when they already match. It only creates when an
authoritative read confirms the object is absent, and only updates when something actually
differs. A bounded retry (`k8s.io/client-go/util/retry`, `retry.OnError` with `retry.DefaultBackoff`)
handles genuine races - `AlreadyExists` from a concurrent create, or `Conflict` from a concurrent
update - by re-reading and retrying the compare/write with a fresh object.

User-visible behavior is unchanged: the object still ends up in the same desired state. The benefit
is narrower - fewer redundant write requests against the API server when nothing has changed.

Fixes #

Validation performed locally (no live cluster available in this environment):
- `go build ./...` passes.
- `go vet ./pkg/secrets-store/...` is clean.
- `go test ./pkg/secrets-store/... -run TestCreateOrUpdateSecretProviderClassPodStatus -v` passes,
  including a new regression test,
  `TestCreateOrUpdateSecretProviderClassPodStatus_NoOpWhenUpToDate`, which seeds a fake client with
  an object that already matches the desired node label, status, and owner references, and wires
  `interceptor.Funcs` to fail the test if `Create` or `Update` is ever called. This test FAILS
  against the pre-fix code (both an unexpected `Create` attempt on `AlreadyExists` and an
  unexpected `Update` call fire) and PASSES against the post-fix code, which is the concrete
  reproduction of the bug described in the issue.
- The pre-existing `TestCreateOrUpdateSecretProviderClassPodStatus` create/update-different-node-label
  subtests still pass unmodified, confirming genuinely new or changed objects still take the
  create/update path correctly.
- Not run: a live cluster / e2e reproduction of the exact `POST 409` + `PUT` API server traffic
  pattern described in the issue. The change is a pure controller-logic fix validated with a
  targeted unit test that reproduces the defect (unconditional write on an unchanged object) and
  proves the fix; it does not depend on cluster or provider-specific runtime behavior.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

Fixes #2058

```release-note
fix: make SecretProviderClassPodStatus reconciliation idempotent
```

### DIFF
--- a/pkg/secrets-store/utils.go
+++ b/pkg/secrets-store/utils.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -98,10 +100,10 @@ func getSecretProviderItem(ctx context.Context, c client.Client, name, namespace
 }
 
 // createOrUpdateSecretProviderClassPodStatus creates secret provider class pod status if not exists.
-// if the secret provider class pod status already exists, it'll update the status and owner references.
+// if the secret provider class pod status already exists but its node label, status or owner references
+// are out of date, it'll be updated. If it already matches the desired state, no write is made.
 func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Client, reader client.Reader, podname, namespace, podUID, spcName, targetPath, nodeID string, mounted bool, objects map[string]string) error {
 	var o []secretsstorev1.SecretProviderClassObject
-	var err error
 	spcpsName := podname + "-" + namespace + "-" + spcName
 
 	for k, v := range objects {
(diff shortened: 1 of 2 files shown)


pujitha24:
/assign @aramase

This is ready for review whenever you have a chance — happy to make any adjustments.

PR #2062, opened 2026-08-19
fix: skip no-op Update in createOrUpdateK8sSecret

jiashun0011:
**What type of PR is this?**
/kind bug

Fixes a regression introduced in v1.6.0 where the synced Kubernetes Secret is re-written on every kubelet republish cycle even when the provider content has not changed.

Since v1.6.0, `CSIDriver.spec.requiresRepublish` is set to `true` and the dedicated rotation reconciler was removed in favor of kubelet-driven `NodePublishVolume` republish.
`SecretProviderClassPodStatusReconciler.createOrUpdateK8sSecret` calls `client.Update()` unconditionally on every reconcile. Because the desired Secret is constructed fresh (no `ResourceVersion`, no
server-side fields), each Update causes the API server to record a new `managedFields` entry for the `csi-secrets-store` field manager, bump `resourceVersion`, and emit a watch event — even when `data`,
`labels`, `annotations`, and `type` are byte-identical to what is already stored.

Downstream controllers doing generic controller-runtime watches on the synced Secret (e.g., operators that reconcile StatefulSets on Secret update events) treat every republish cycle as a real rotation and
take disruptive action (pod recreation, rolling restart, etc.) even when nothing in the external secret store has changed.

This PR:
- Fetches the existing Secret before calling `Update()` and skips the write entirely when `type`, `data`, `labels`, and `annotations` already match.
- When an Update is still required, propagates the fetched `ResourceVersion` so the write uses proper optimistic concurrency instead of an unconditional overwrite.
- Adds a `secretContentEqual` helper covering the fields this controller manages.

Reproduction (v1.6.0-6 driver, v1.8.2-5 Azure provider, k8s 1.35, `enable-secret-rotation=true`, `rotation-poll-interval=2m`): synced Secret `resourceVersion` and `managedFields[csi-secrets-store].time`
advance every ~5 minutes while the Key Vault object version and content are unchanged. With this fix applied, the synced Secret stays at the same `resourceVersion` across cycles.

Fixes #

- Behavior parity with v1.5.x: v1.5.x invoked `createOrUpdateK8sSecret` only after a content-diff check inside the rotation reconciler, so no-op writes never reached the API server. Since v1.6.0 relies on
kubelet republish + unconditional Update, that diff needs to move into `createOrUpdateK8sSecret` itself; this PR does exactly that.
- Optional follow-up (not in this PR): switch the write to Server-Side Apply with a stable field manager. That would let the API server drop no-op applies natively, but is a larger change; the conservative
Get-then-compare approach here is enough to restore the pre-v1.6.0 contract.
- Reloader-style consumers (data-hash based) are unaffected either way.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


stlaz:
/close
duplicate

PR #2065, opened 2026-08-27
Fix secret update hotloop

pmengelbert:
**What type of PR is this?**

/kind bug
fix: prevent secret update hotloop

The driver rewrote each managed secret on every reconcile, which kept triggering fresh reconciles for the same object. Now it fetches the existing secret first and skips the write entirely when the data hasn't changed.

When the data does change, apply it with a merge patch that only touches the Data field instead of doing a full Update. That leaves owner references, labels, and annotations in place.

Break out of the create/update backoff early on a conflict or already-exists error to enter the next reconcile more quickly.

Add tests covering the hotloop case, the data-only patch, and the fact that updates don't affect labels, annotations, or owner references.

Addresses issue 2057.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controllers/secretproviderclasspodstatus_controller.go
+++ b/controllers/secretproviderclasspodstatus_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
@@ -34,6 +36,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -296,7 +299,6 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context, 
 			continue
 		}
 
-		var funcs []func() (bool, error)
 		secretType := secretutil.GetSecretType(strings.TrimSpace(secretObj.Type))
 
 		var datamap map[string][]byte
@@ -320,9 +322,19 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context, 
 		// only on secrets created and managed by the driver
 		labelsMap[SecretManagedLabel] = "true"
 
-		createFn := func() (bool, error) {
+		if err := wait.ExponentialBackoff(wait.Backoff{
+			Steps:    5,
+			Duration: 1 * time.Millisecond,
+			Factor:   1.0,
+			Jitter:   0.1,
+		}, func() (done bool, err error) {
 			if err := r.createOrUpdateK8sSecret(ctx, secretName, req.Namespace, datamap, labelsMap, annotationsMap, secretType); err != nil {
 				klog.ErrorS(err, "failed to create Kubernetes secret", "spc", klog.KObj(spc), "pod", klog.KObj(pod), "secret", klog.ObjectRef{Namespace: req.Namespace, Name: secretName}, "spcps", klog.KObj(spcPodStatus))
+				// error out here to break out of the backoff and retry the full Reconcile() early
+				if apierrors.IsConflict(err) || apierrors.IsAlreadyExists(err) {
+					return false, err
+				}
+
 				// syncSecret.enabled is set to false by default in the helm chart for installing the driver in v0.0.23+
 				// that would result in a forbidden error, so generate a warning that can be helpful for debugging
 				if apierrors.IsForbidden(err) {
@@ -331,19 +343,9 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context, 
 				return false, nil
 			}
 			return true, nil
-		}
-		funcs = append(funcs, createFn)
-
-		for _, f := range funcs {
-			if err := wait.ExponentialBackoff(wait.Backoff{
-				Steps:    5,
-				Duration: 1 * time.Millisecond,
-				Factor:   1.0,
-				Jitter:   0.1,
-			}, f); err != nil {
-				r.generateEvent(pod, corev1.EventTypeWarning, secretCreationFailedReason, err.Error())
-				return ctrl.Result{RequeueAfter: 5 * time.Second}, err
-			}
+		}); err != nil {
+			r.generateEvent(pod, corev1.EventTypeWarning, secretCreationFailedReason, err.Error())
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, err
 		}
 	}
 
@@ -399,31 +401,63 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta 
 // createOrUpdateK8sSecret creates K8s secret with data from mounted files
 // If a secret with the same name already exists in the namespace of the pod, it will update that existing secret.
 func (r *SecretProviderClassPodStatusReconciler) createOrUpdateK8sSecret(ctx context.Context, name, namespace string, datamap map[string][]byte, labelsmap map[string]string, annotationsmap map[string]string, secretType corev1.SecretType) error {
-	secret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace:   namespace,
-			Name:        name,
-			Labels:      labelsmap,
-			Annotations: annotationsmap,
-		},
-		Type: secretType,
-		Data: datamap,
+	secret := &corev1.Secret{}
+	getErr := r.reader.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, secret)
+
+	// Secret does not exist, create it
+	if apierrors.IsNotFound(getErr) {
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:   namespace,
+				Name:        name,
+				Labels:      labelsmap,
+				Annotations: annotationsmap,
+			},
+			Type: secretType,
+			Data: datamap,
+		}
+
+		err := r.writer.Create(ctx, secret)
+		if err == nil {
+			klog.InfoS("successfully created Kubernetes secret", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
+		}
+
+		return err
 	}
 
-	err := r.writer.Create(ctx, secret)
-	if err == nil {
-		klog.InfoS("successfully created Kubernetes secret", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
+	if getErr != nil {
+		return getErr
+	}
+
+	// Secret exists, update it
+	klog.V(5).InfoS("Kubernetes secret is already created", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
+
+	if reflect.DeepEqual(secret.Data, datamap) {
 		return nil
 	}
-	if !apierrors.IsAlreadyExists(err) {
+
+	oldData, err := json.Marshal(secret)
+	if err != nil {
+		return fmt.Errorf("failed to marshal old secret, err: %w", err)
+	}
+
+	secret.Data = datamap
+
+	newData, err := json.Marshal(secret)
+	if err != nil {
+		return fmt.Errorf("failed to marshal new secret, err: %w", err)
+	}
+
+	// Patching data clobbers existing data
+	patch, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, secret)
+	if err != nil {
+		return fmt.Errorf("failed to create patch, err: %w", err)
+	}
+
+	if err = r.writer.Patch(ctx, secret, client.RawPatch(types.MergePatchType, patch)); err != nil {
 		return err
 	}
 
-	klog.V(5).InfoS("Kubernetes secret is already created", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
-	err = r.writer.Update(ctx, secret)
-	if err != nil {
-		return err
-	}
 	klog.V(5).InfoS("successfully updated Kubernetes secret", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
 	return nil
 }
--- a/controllers/secretproviderclasspodstatus_controller_test.go
+++ b/controllers/secretproviderclasspodstatus_controller_test.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 )
 
 var (
@@ -121,6 +122,13 @@ func newReconciler(client client.Client, scheme *runtime.Scheme, nodeID string) 
 	}
 }
 
+// expectOwnerRefs asserts the secret's owner references match want exactly, by
+// full-struct equality and independent of order. ConsistOf uses DeepEqual, so
+// pointer fields (Controller, BlockOwnerDeletion) are compared by their pointees.
+func expectOwnerRefs(g Gomega, secret *corev1.Secret, want ...metav1.OwnerReference) {
+	g.Expect(secret.OwnerReferences).To(ConsistOf(want))
+}
+
 func TestPatchSecretWithOwnerRef(t *testing.T) {
 	g := NewWithT(t)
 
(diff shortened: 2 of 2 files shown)


benjaminapetersen:
Need to add to at least to `eedp2e` (3rd commit):

```
Co-authored-by: Benjamin A. Petersen <admin@benjaminapetersen.me>
```
Thanks!

Will take a 👀  on the rest.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +400,68 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+	}
+
+	// Secret exists, update it
 	klog.V(5).InfoS("Kubernetes secret is already created", "secret", klog.ObjectRef{Namespace: namespace, Name: name})
-	err = r.writer.Update(ctx, secret)
-	if err != nil {
+
+	// The return from a controller-runtime `Get` operation does not guarantee
+	// ownership (for example, it could be a memory reference to a cache entry),
+	// so have to be copied before they can be modified.
```
stlaz: I don't think this is true. Getting a deep copy from Gets is a well known drawback of controller runtime.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +400,68 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+	secret.Type = secretType
+
+	if equality.Semantic.DeepEqual(orig, secret) {
+		return nil
+	}
+
+	if err := r.writer.Update(ctx, secret); err != nil {
+		// If the object has been updated by another caller, we do not know the
+		// proper authority. In that case, the update request shouldn't be
+		// re-queued.
```
stlaz: I don't understand this. We should retry until the object has the shape we expect it to?

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +195,86 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
...
+		}).
+		Build()
+
+	reconciler := newReconciler(client, scheme, "node1")
+	g.Expect(err).NotTo(HaveOccurred())
+	err = reconciler.Patcher(context.TODO())
+	secret := &corev1.Secret{}
+	err = client.Get(context.TODO(), types.NamespacedName{Name: "secret1", Namespace: "default"}, secret)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(err).NotTo(HaveOccurred())
```
stlaz, line 270: de-duplicate

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +195,86 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
...
+	g.Expect(len(secret.OwnerReferences)).To(Equal(1))
+	g.Expect(secret.OwnerReferences[0].APIVersion).To(Equal("apps/v1"))
+	g.Expect(secret.OwnerReferences[0].Kind).To(Equal("ReplicaSet"))
+	g.Expect(secret.OwnerReferences[0].Name).To(Equal("pod-6886c65f8f"))
+	g.Expect(secret.OwnerReferences[0].UID).To(Equal(types.UID("f39da13d-7246-4ef5-aed4-a6905f82cbcd")))
```
stlaz: notice you're not checking all the fields of the owner ref.

I think this might work, or you might need to use something like semantic DeepEqual check
```suggestion
    g.Expect(secret.OwnerReferences[0]).To(Equal(metav1.OwnerReference{
        APIVersion: ...
        BlockOwnerDeletion: ...
        ...
    }))
```

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +195,86 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
...
+
+	// secret already exists, just update it.
+	err = reconciler.createOrUpdateK8sSecret(context.TODO(), "secret1", "default", map[string][]byte{"foo": []byte("bar")}, labels, annotations, corev1.SecretTypeOpaque)
```
stlaz, line 263: If it's possible to mock the expected contents of the secret, it would be safer to run the whole Reconcile() method for the sake of this test.

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +195,86 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
...
+	err = reconciler.Patcher(context.TODO())
+	g.Expect(err).NotTo(HaveOccurred())
+
+	g.Expect(createCount).To(Equal(0))
+	g.Expect(updateCount).To(Equal(1))
+	g.Expect(patchCount).To(Equal(1))
+
+	g.Expect(secret.Labels).To(Equal(labels))
+
+	g.Expect(secret.Name).To(Equal("secret1"))
```
stlaz: We don't need to check this, but you may want to check the content of .Data and annotations?

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +195,86 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
 	g.Expect(secret.Name).To(Equal("my-secret2"))
 }
 
+func TestCreateOrUpdateHotloop(t *testing.T) {
```
stlaz, line 199: I would expect this test scenario to be:

1. Reconcile() or createOrUpdate() creates the secret
2. Check create=1, update=0, patch=0
3. Patch() adds the owner refs
4. check create=1, update=0, patch=1, check correct owner refs
5. Reconcile() or createOrUpdate() updates the secret with new mocked content
6. check create=1, update=1, patch=1, check correct owner refs
7. Patch() does nothing
8. check create=1, update=1, patch=1, check correct owner refs

to really make sure that we're not hotlooping on the secret we created ourselves, which is likely to be what happens in the real world.

stlaz (changes requested):

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: If the code hits this path, then someone else has updated the object between our `Get` and the update we have submitted a request for. The "update" is based on the assumption that the object has not been changed by someone else. What is supposed to happen here if a user updates a secret using `kubectl` between the `Get` and `Update` calls?

IIUC when the next reconcile is triggered, then `createOrUpdateK8sSecret` will be called again and the secret will be reconciled to the expected shape. Please clarify if I am mistaken.

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: Got it. Also now that I think about it, it would be impossible anyway, given the way the code is structured. The pointer to `secret` is an input to the `Get` call and it's not possible for `Get` to change the address contained in the pointer. The memory is stack or heap allocated by us and controlled by this function.

controllers/secretproviderclasspodstatus_controller_test.go
pmengelbert: Is it necessary to check all the fields? This was adapted from `TestPatcherForPodWithOwner`. I've added `Controller` and `BlockOwnerDeletion` which were the missing fields relative to the struct built out in the `initObjects` struct.

controllers/secretproviderclasspodstatus_controller.go
stlaz: > What is supposed to happen here if a user updates a secret using kubectl between the Get and Update calls?

we need to stomp the user change in favour of whatever we think is the correct shape of the secret

> IIUC when the next reconcile is triggered, then createOrUpdateK8sSecret will be called again and the secret will be reconciled to the expected shape.

right, by returning an error here we would make sure that the next reconcile is triggered

controllers/secretproviderclasspodstatus_controller_test.go
stlaz: I wouldn't stick with the old code too much.

I'd rather check all the fields to make sure the patched version is correct. You can also fix `TestPatcherForPodWithOwner` if you want, and have a common helper for both these tests.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -87,7 +88,7 @@ func New(driverName string, mgr manager.Manager, nodeID string) (*SecretProvider
 }
 
 func (r *SecretProviderClassPodStatusReconciler) RunPatcher(ctx context.Context) {
-	ticker := time.NewTicker(5 * time.Second)
+	ticker := time.NewTicker(1 * time.Second)
```
dwickr: Is this ticker change related to the fix?

stlaz: I also think the original timing was better.

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: After peeking at the controller runtime code, the reader can be configured to deep copy or not. I'm not sure which is happening here, but it's probably simplest to just do the deep copy. Additionally, while we own `secret` here, its fields could be set to pointers to memory owned by someone else (thanks to Mo for pointing this out).  I'll update the comment to be more accurate.

The other option is to assert or check that the system is configured to return a deep copy, but I don't know that it's worth it over doing a deep copy.

stlaz: We don't set that option. You can remove the deep copy.

controllers/secretproviderclasspodstatus_controller.go
stlaz: @pmengelbert please revert

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: @stlaz we also need to propagate the error up through the `ConditionFunc`.

@enj if you are wondering why it changed, check the commit message for more detail.

controllers/secretproviderclasspodstatus_controller_test.go
pmengelbert: done

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: I believe this is a typo in the first place.

pmengelbert: fixed.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +401,71 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+
+	// The return from a controller-runtime `Get` operation does not guarantee
+	// ownership (for example, it could be a memory reference to a cache entry),
+	// so have to be copied before they can be modified.
+	unmodified := secret.DeepCopy()
+	secret.Data = datamap
+	secret.Type = secretType
+
+	// Merge (not replace) labels and annotations so as to not compete with other controllers
+	if secret.Labels == nil && len(labelsmap) != 0 {
```
stlaz: IIRC these would be rewritten previously? Keep the old behavior then, we don't want more changes to the behavior than necessary.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +401,71 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+		secret.Labels = make(map[string]string, len(labelsmap))
+	}
+	for k, v := range labelsmap {
+		secret.Labels[k] = v
+	}
```
stlaz: nit: this feels a little easier on my eyes but could be just personal preference

Also note `maps.Copy()`, a somewhat new helper function.
```suggestion
    if len(labelsmap) != 0 {
	    // Merge (not replace) labels and annotations so as to not compete with other controllers
    	if secret.Labels == nil {
	    	secret.Labels = make(map[string]string, len(labelsmap))
	    }
        maps.Copy(secret.Labels, labelsmap)
    }
```

controllers/secretproviderclasspodstatus_controller.go
stlaz, line 342: note for future ourselves - we should not swallow _any_ errors here.

For now I don't care enough to want to have this fixed in this PR, but would be a great follow-up. The fix shouldn't be naive - we must not log the error every time, but should log each error string exactly once.

stlaz (approved):
/lgtm
/approve

Look at the suggestions and consider a follow-up with error logging during createOrUpdate.

You will also need to fix the gofmt issue. Fix your IDE to run it on save.

stlaz:
/lgtm

/test pull-secrets-store-csi-driver-e2e-windows
/test pull-secrets-store-csi-driver-e2e-azure
looks like infra issues

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +402,72 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
-	err = r.writer.Update(ctx, secret)
-	if err != nil {
+
+	// The return from a controller-runtime `Get` operation does not guarantee
+	// ownership (for example, it could be a memory reference to a cache entry),
+	// so have to be copied before they can be modified.
+	unmodified := secret.DeepCopy()
+	secret.Data = datamap
+	secret.Type = secretType
+
```
enj, line 438: Comment doesn't line up with the code anymore.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +402,72 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+	if len(labelsmap) != 0 {
+		// Merge (not replace) labels and annotations so as to not compete with other controllers
+		if secret.Labels == nil {
+			secret.Labels = make(map[string]string, len(labelsmap))
+		}
+		maps.Copy(secret.Labels, labelsmap)
+	}
+
+	if len(annotationsmap) != 0 {
+		// Merge (not replace) annotations and annotations so as to not compete with other controllers
```
enj: If you merge, how does the user ever drop labels from the secret?

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +402,72 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+				Annotations: annotationsmap,
+			},
+			Type: secretType,
+			Data: datamap,
+		}
+
+		err := r.writer.Create(ctx, secret)
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
```
enj: Seems like you would want to let this error cause a retry to be consistent with the conflict handling above?

enj (changes requested):
First pass.

controllers/secretproviderclasspodstatus_controller.go
stlaz: +1

controllers/secretproviderclasspodstatus_controller.go
stlaz: Per mtg: just leave the labels/annotations as they are in the secret

edit: just copy-paste this - https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/fcfb81f91403e935e6958989805c974b34d25187/pkg/rotation/reconciler.go#L568-L587

controllers/secretproviderclasspodstatus_controller.go
benjaminapetersen: Great, will return the err, agree.

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -177,15 +179,15 @@ func (r *SecretProviderClassPodStatusReconciler) Patcher(ctx context.Context) er
...
+				if apierrors.IsConflict(err) || apierrors.IsTimeout(err) {
+					return false, nil
 				}
 				// syncSecret.enabled is set to false by default in the helm chart for installing the driver in v0.0.23+
 				// that would result in a forbidden error, so generate a warning that can be helpful for debugging
 				if apierrors.IsForbidden(err) {
 					klog.Warning(SyncSecretForbiddenWarning)
 				}
-				return false, nil
+				return false, err
```
stlaz: Why this change? It seems worse, except maybe for the forbidden case where this is a behavior change, but we probably don't need to worry about it right now?

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -320,9 +321,21 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context,
...
+			Steps:    5,
+			Duration: 1 * time.Millisecond,
+			Factor:   1.0,
+			Jitter:   0.1,
+		}, func() (done bool, err error) {
 			if err := r.createOrUpdateK8sSecret(ctx, secretName, req.Namespace, datamap, labelsMap, annotationsMap, secretType); err != nil {
 				klog.ErrorS(err, "failed to create Kubernetes secret", "spc", klog.KObj(spc), "pod", klog.KObj(pod), "secret", klog.ObjectRef{Namespace: req.Namespace, Name: secretName}, "spcps", klog.KObj(spcPodStatus))
+				// Returning the error here short-circuits the remaining
+				// ExponentialBackoff and triggers a full reconcile in 5
+				// seconds (vs 5 minutes for the next timed reconcile).
```
stlaz: Feels AI-ey, I prefer concise, ideally one-line comments to avoid cognitive overload:
`error out here to break out of the backoff and retry the full Reconcile() early`

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -320,9 +321,21 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context,
...
+				if apierrors.IsConflict(err) {
```
stlaz: you forgot to handle `IsAlreadyExists()`

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +402,59 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+	}
+	newData, err := json.Marshal(&newSecret)
+	if err != nil {
+		return fmt.Errorf("failed to marshal new secret, err: %w", err)
+	}
+	if bytes.Equal(oldData, newData) {
+		return nil
+	}
+
+	if err := r.writer.Update(ctx, &newSecret); err != nil {
```
stlaz: we agreed on Patch()-ing with the copy-pasted code

controllers/secretproviderclasspodstatus_controller_test.go
```diff
@@ -188,6 +202,132 @@ func TestCreateOrUpdateK8sSecret(t *testing.T) {
...
+
+	// Simulate another controller adding metadata and changing the type.
+	secret := getSecret()
+	secret.Labels["foreign.example.com/team"] = "x"
+	secret.Annotations["foreign.example.com/owned"] = "yes"
+	secret.Type = corev1.SecretTypeTLS
+	err = client.Update(context.TODO(), secret)
+	g.Expect(err).NotTo(HaveOccurred())
+	updateCount = 0
+
```
stlaz: this does not belong here, you broke the original workflow and all of the numbered comments now don't make sense

stlaz (changes requested):

controllers/secretproviderclasspodstatus_controller.go
pmengelbert: @stlaz I definitely wrote that comment by hand and formatted it with vim. i'm fine with making it more concise, but I resent the implication!

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +405,70 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+
+	if reflect.DeepEqual(secret.Data, datamap) {
 		return nil
 	}
-	if !apierrors.IsAlreadyExists(err) {
-		return err
+
+	// The return from a controller-runtime `Get` operation does not guarantee
+	// ownership (for example, it could be a memory reference to a cache entry),
+	// so have to be copied before they can be modified.
```
enj:
```suggestion
```

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -399,31 +405,70 @@ func (r *SecretProviderClassPodStatusReconciler) processIfBelongsToNode(objMeta
...
+			},
+			Type: secretType,
+			Data: datamap,
+		}
+
+		err := r.writer.Create(ctx, secret)
+		if apierrors.IsAlreadyExists(err) {
+			return err
+		}
+
```
enj:
```suggestion
```

controllers/secretproviderclasspodstatus_controller.go
```diff
@@ -320,9 +322,23 @@ func (r *SecretProviderClassPodStatusReconciler) Reconcile(ctx context.Context,
...
+		}, func() (done bool, err error) {
 			if err := r.createOrUpdateK8sSecret(ctx, secretName, req.Namespace, datamap, labelsMap, annotationsMap, secretType); err != nil {
 				klog.ErrorS(err, "failed to create Kubernetes secret", "spc", klog.KObj(spc), "pod", klog.KObj(pod), "secret", klog.ObjectRef{Namespace: req.Namespace, Name: secretName}, "spcps", klog.KObj(spcPodStatus))
+				// error out here to break out of the backoff and retry the full Reconcile() early
+				if apierrors.IsConflict(err) {
+					return false, err
+				}
+
+				if apierrors.IsAlreadyExists(err) {
+					return false, err
```
enj: Combine.

enj (changes requested):
Second pass.

benjaminapetersen:
Addresses #2057 
For a more explicit cross-reference.

PR #2092, opened 2026-09-14
chore: trigger 1.5.7 image build

pmengelbert:


benjaminapetersen:
/ok-to-test

Not sure why @pmengelbert isn't in the "already ok to test" list. 🤔

PR #2096, opened 2026-09-19
fix: make SecretProviderClassPodStatus reconciliation idempotent

### LINKED ISSUE #2058: SecretProviderClassPodStatus reconciliation unconditionally creates and updates existing objects

goutamadwant:
**What type of PR is this?**

/kind bug

Reads the `SecretProviderClassPodStatus` before writing it, creates it only when it does not exist, and skips the update when the node label, status, and owner reference are already current.

Create races and update conflicts are retried with an authoritative API read so a stale filtered cache cannot keep the retry loop on an obsolete object.

**Which issue(s) this PR fixes**:

Fixes #2058

This replaces the approach from closed PR #2060 and adds coverage for the no-op path, create races, update conflicts, and filtered-cache misses.

Validation:

- Focused reconciliation regression tests pass. The unchanged-object regression fails against the original implementation because it performs both a create and an update.
- `go test ./... -skip '^(TestEnsureMountPoint|TestSanity)$' -count=1` passes.
- `make sanity-test` passes: 12 passed, 66 skipped by the project configuration and capability checks.
- `go vet ./pkg/secrets-store/...` passes.

Validation ran on macOS using a short temporary directory for Unix sockets. `TestEnsureMountPoint` is excluded because its fake mount is not observed as a real mount on this host. Live Kubernetes/provider end-to-end tests have not been run.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


pkg/secrets-store/utils.go
```diff
@@ -132,30 +133,52 @@ func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Cl
...
-		if !apierrors.IsNotFound(err) {
+	key := client.ObjectKey{Name: spcpsName, Namespace: namespace}
+	readFromAPI := false
+	return retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return apierrors.IsAlreadyExists(err) || apierrors.IsConflict(err)
+	}, func() error {
+		spcps := &secretsstorev1.SecretProviderClassPodStatus{}
+		var err error
+		if readFromAPI {
+			err = reader.Get(ctx, key, spcps)
```
stlaz: remove the live lookup branch, rely on informer resync

pkg/secrets-store/utils.go
```diff
@@ -132,30 +133,52 @@ func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Cl
 		},
 	})
 
-	if err = c.Create(ctx, spcPodStatus); err == nil || !apierrors.IsAlreadyExists(err) {
-		return err
-	}
-	klog.InfoS("secret provider class pod status already exists, updating it", "spcps", klog.ObjectRef{Name: spcPodStatus.Name, Namespace: spcPodStatus.Namespace})
```
stlaz: this part is fine, leave as is

pkg/secrets-store/utils.go
```diff
@@ -132,30 +133,52 @@ func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Cl
...
-
-	spcps := &secretsstorev1.SecretProviderClassPodStatus{}
-	// the secret provider class pod status with the name already exists, update it
-	if err = c.Get(ctx, client.ObjectKey{Name: spcpsName, Namespace: namespace}, spcps); err != nil {
-		if !apierrors.IsNotFound(err) {
+	key := client.ObjectKey{Name: spcpsName, Namespace: namespace}
+	readFromAPI := false
```
stlaz: remove this please

pkg/secrets-store/utils.go
```diff
@@ -132,30 +133,52 @@ func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Cl
...
+			err = c.Create(ctx, spcPodStatus.DeepCopy())
+			if apierrors.IsAlreadyExists(err) {
+				readFromAPI = true
+			}
 			return err
 		}
-		// the secret provider class pod status could be missing in the cache because it was labeled with a different node
-		// label, so we need to get it from the API server
-		if err = reader.Get(ctx, client.ObjectKey{Name: spcpsName, Namespace: namespace}, spcps); err != nil {
+		if err != nil {
```
stlaz, line 159: let's split the creation and updates
```suggestion
    podStatusKey := client.ObjectKey{Name: spcpsName, Namespace: namespace}
    if err := c.Get(ctx, podStatusKey, spcPodStatus); apierrors.IsNotFound(err) {
        if createErr := c.Create(ctx, spcPodStatus); apierrors.IsAlreadyExists(createErr) {
            // TODO: log the error here
            // force live fetch in update loop below
            spcPodStatus = nil
        }
        return err
    } else if err != nil {
        return err
    }
    
    return retry.RetryOnConflict(retry.DefaultRetry, func(err error) bool {
        defer func() { spcPodStatus = nil }
        if spcPodStatus == nil {
            spcPodStatus = &secretsstorev1.SecretProviderClassPodStatus{}
            // live get
            if err := reader.Get(ctx, podStatusKey, spcPodStatus); err != nil {
                return err
            }
        }
        // update the object here
    }
```

pkg/secrets-store/utils_test.go
```diff
@@ -237,6 +258,222 @@ func TestCreateOrUpdateSecretProviderClassPodStatus(t *testing.T) {
...
+				t.Errorf("Create calls = %d, want %d", createCount, tt.wantCreates)
+			}
+			if updateCount != tt.wantUpdates {
+				t.Errorf("Update calls = %d, want %d", updateCount, tt.wantUpdates)
+			}
 		})
 	}
 }
+
+func TestCreateOrUpdateSecretProviderClassPodStatusSkipsUnchangedObject(t *testing.T) {
```
stlaz, line 271: it feels like we should be able to turn these into table tests?

pkg/secrets-store/utils.go
```diff
@@ -132,30 +133,52 @@ func createOrUpdateSecretProviderClassPodStatus(ctx context.Context, c client.Cl
...
+			spcps.Labels = make(map[string]string)
+		}
+		spcps.Labels[secretsstorev1.InternalNodeLabel] = nodeID
+		spcps.Status = spcPodStatus.Status
+		spcps.OwnerReferences = spcPodStatus.OwnerReferences
+		err = c.Update(ctx, spcps)
+		if apierrors.IsConflict(err) {
+			readFromAPI = true
+		}
+		return err
```
stlaz, line 180: I'd extract the changes and also apply it to the object to be created couple lines above
```go
func updateSecretProviderClassStatus(spcps *secretsstorev1.SecretProviderClassPodStatus, nodeID string, podStatus secretsstorev1.SecretProviderClassPodStatusStatus, ownerRefs []metav1.OwnerReference) bool {
    // TODO: change the object if needs be, return if changed
}
```

stlaz (commented):
Thank you for your PR!

I'd restructure the code a little bit, but generally looks good.

